PR #41, opened 2014-03-02
Initial version of the same-origin policy model in Alloy

eskang:
This initial version contains Alloy models of:
* client-server model with HTTP 
* the same origin policy
* the cross-origin sharing 
* security property

To add for the future versions:
* model documentation
* JSONP
* PostMessage


MichaelDiBernardo:
Merging this in preparation for copy-editing. Thank you!

PR #42, opened 2014-03-03
Add lexer and parser for Python 3

hausdorff:
The necessities for building, like the `.cabal` file and `Makefile` are in the root directory (i.e., `parser-and-lexer/`).

The code under submission is in `src/`, and the Haskell code that supports the main submission (e.g., code to interface with libraries) is under `src/lib/`. The directory `src/tools/` contains miscellaneous tools written in whatever language I felt like (e.g., they pretty-print the AST) and can be safely ignored; though the `Makefile` uses them to print nice things for the user, they are purely cosmetic.

The code itself is still a little overweight, but is approaching what it should look like. The parser is in `src/Parser.hs`, and clocks in at nearly 700 lines (this will shrink soon). The lexer is in `src/Lexer.hs`, and clocks in at just over 500 lines, but will be easily compressed. What work really remains to be done is to compress `src/LexerUtils.hs`, which contains all the regexes used to lex the Python. There additionally are `src/PowLex.hs` and `src/PowParse.hs`, which are both < 10 line files that contain "main" methods for the lexer and parser respectively. At this point they too are under submission, though there isn't anything interesting going on in there.

TO REVIEWERS: I recommend you start with the parser first, because it has an accompanying [http://blog.nullspace.io/obvious-python-parser.html](blog post) right now, while the lexer does not. It is also much closer to the final product, with great pains taken to make it as readable as possible.


MichaelDiBernardo:

PR #43, opened 2014-03-03
The Python interpreter

akaptur:
The Python interpreter draft.  

Other comments:
- I've checked in tests, although they're not part of the 500 lines / not for inclusion in the book.  (They're not that interesting anyway - just verifying that little snippets of Python do the right thing.)
- It was less work than I thought it would be to adapt this from @nedbat's byterun, so we should make sure he gets appropriate credit.


akaptur:
@MichaelDiBernardo, this is pretty much where I'd like to end up! Thanks for your endurance :)

MichaelDiBernardo:
@akaptur thanks! Will do a final read this week and let you know. Based on the solidity of your previous draft, I imagine we'll be good to merge.

Also, I may beg you to let us steal 'powerful programs written in 500 lines or less' as our motto for this volume :)

MichaelDiBernardo:
@akaptur: Terrific. Merging! We'll be in touch when we're done copy-editing to get your feedback on our changes.

PR #45, opened 2014-03-04
Python bytecode compiler

darius:
This is still quite a rough draft. I'd always welcome feedback, but expect it to change plenty over the next week, and if you have limited time it'd be best to wait till then to spend most of it.

The 'real' code is in {codegen,assembler,scoper}.py. Scoper currently calls on the built-in Python compiler's symtable module for its scope analysis, which is sort of cheating; this is meant as scaffolding to avoid a class of bugs while refining the rest of the compiler, and I hope to come back to it. The immediate next jobs are a test suite, filling in some details it's still sloppy about (so that it's clear just what Python subset is supported, and it complains clearly for input outside that subset), and supporting some more of Python: the big missing feature is nested functions and lambda expressions. (That might turn out too much for the complexity budget.)

### DIFF
--- a/bytecode-compiler/Makefile
+++ b/bytecode-compiler/Makefile
@@ -1,9 +1,9 @@
 all: tailbiter0.py tailbiter1.py tailbiter2.py
 
-doc: ~/Desktop/ch1.pdf
+doc: ~/Desktop/tailbiter.pdf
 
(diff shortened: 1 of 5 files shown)


MichaelDiBernardo:

PR #63, opened 2014-06-02
First draft of typesetting engine

jilljenn:
I implemented the Knuth-Plass algorithm. If necessary, I can simplify the code (the current version may work for circle or triangle-shaped text, [see this example](http://www.bramstein.com/projects/typeset/) but we don't need such advanced use).

The PostScript part still needs to be done.


MichaelDiBernardo:

PR #80, opened 2014-08-21
Blockcode chapter outline

dethe:
I've updated the README.md based on Debo's feedback, specifically:

* Introduce the concept of block-based languages
* Talk more about why the files are broken down the way they are, and more about why this architecture was selected generally
* Slow down a bit and provide more introduction and explanation before diving into code comments
* Unlink text and provide the URLs as endnotes (I made up my own convention for this because I don't think Markdown supports it, hopefully it is clear).

I still need to flesh out two sections: "Swapping out turtles" on how the language itself is pluggable, and "Lessons Learned" for the summary. It would be good to have feedback on what is there now to see if I'm heading in the right direction with this.


MichaelDiBernardo:
Hi @dethe, just read this over now. Thanks for adding the code; you did a very good job of spacing things out manageably.

I think that actually showing at least the introductory parts of a second example language (your sprite manipulation language) is a good idea, if it's within your capacity to do so. Do you think it will be roughly the same length + complexity as the turtle example? If it's going to be more complicated, we might reconsider adding it (since it will be a lot to digest near the end), or just showing a subset of the implementation.

In exchange, I would probably truncate the `utils.js` section to simply say "this is a collection of things that we would normally expect the browser to support natively; if you are interested, you can look at the source."

In any case, I think that the basic structure of what you've got here is ready for review. I'll reach out to some reviewers now, since it normally takes 7-14 days before anyone can take a close look at a draft. Does that work for you?

dethe:
Hi @MichaelDiBernardo, sorry I completely missed that you'd replied here. The second example language would be longer (making engaging visual languages in a small amount of code is hard!) but I could show/discuss a subset. Happy to drop the `utils` discussion, but I should probably add some more comments to that file in that case.

I can add an initial discussion of the second language ASAP and work on completing that language and some examples using it after, since that will have less impact on the editing/reviewing. The core of the language is there, but it needs some work before I can get engaging examples written.

MichaelDiBernardo:
Sounds like a great plan @dethe. Thanks :)

chrisseaton:
I really liked the code that this chapter goes with, but I think the text here could do with some more work.

We go from very high level discussion of block languages and web browsers right down into the weeds of `block.js`. We need something in the middle. We need more architecture and less simple description of the code. I'm missing a general description of how the program is represented in memory, and how it is executed. I think you should be talking about trees and comparing to AST interpreters and the visitor pattern - that's how it works, right?

Maybe another way to put it is that this chapter reads like a report on the project, rather than being a piece of distinct prose to explain how it is has been **architectured** and how it works. It explains too much of the 'how' rather than the 'why'.

I think I could understand most of the things you tell me in the text by looking at the code. I'd like to understand more of your thinking of why you designed the program as you did.

#### General notes

I think you could lose the first two sentences when the text is in the context of the book.

I don't know what 'web-native' means. You say you've stuck to simple tools but then why is it relevant that "Modern web browsers are powerful platforms, with a rich set of tools for building great apps"?

Some filenames are written in monospace and others proportional.

You say "it's all standard web stuff". If it's all standard why would I be interested in reading the chapter :) Also saying how basic it is might put off a less experienced programmer to whom it isn't "standard" at all.

"The entry point for a block". What does entry point mean here? Where it starts to run? Or where it's created from? I think it means the latter but most people would assume the former.

Blockcode is sometimes capitalised and sometimes not.

"simply retrieve the child" should be "simply retrieves the child".

Sentence starting with "it always returns" needs a capital letter.

"To retrieve the script from a block blockScript(block) will return a structure" - at this point you're in danger of telling me that `getX` will get X - it's too much of a simple description of the code that I could read for myself.

"When we want to notify the system to run the assembled script during the next frame handler". What is a frame handler? This term hasn't been defined at all.

In `file.js` something has gone wrong and you have prose in monospace and filenames in proportional.

I think you could basically drop the whole section on `utils.js`.

I think the first few sentences of the second paragraph don't give the best introduction. It talks about how block programming is different from lexical programming, without first saying what it is. For example

"A block-based language differs from programming language where you type in code as words because you drag blocks that represent code into place."

Might be better as something like

"In block-based programming languages you write programs by dragging and connecting blocks that represent parts of the programming. Block-based languages differ from more conventional programming languages where you type words and symbols."

dethe:
Great feedback, @chrisseaton! Let me see what I can do do get that more architectural view of the project.

sevas:
Hey there, I'll be you second reviewer.


So, I think my general comment pretty is pretty much in line with what @chrisseaton said.

I'd like a bit more about the big picture of this application, to better connect the pieces before go through each js file. Something like: 
- `block.js`is the abstract representation of a block-based language
- `index.html`, `drag.js`, and `file.js` implement the editor UI for a block-based language, using the html DOM to represent the AST of a program
- `menu.js` is a script runner for a block-based language
- `turtle.js` implements the semantics of one specific block-based language and its interpreter

I just used AST in the above paragraph, because I also thinks this chapter skims over elements of programming language theory and interpretation without really mentioning it.
I mean, we are building an editor and an interpreter for an AST, it's a good opportunity to make the connection with more "traditional", text-based languages. It's an interesting fact, for instance, that block-based languages don't require lexing and parsing phases, but could share other phases present in compilers and interpreters. For instance, this chapter could mention that it would be possible to augment the interpreter in ``turtle.js`` with different backends, or a byte code generator for another vm, or even a c++ source file. 


### Comments

#### General note on html/js

I've found that the way html/js applications work often confuses programmers who are used to more traditional programming environments. The principal issue is that there is no clear entry point, no `main()` function.
Although, this is not specific to this chapter, but for all chapters that use html/javascript as the implementation language. Maybe one chapter should take care of mentioning the magic that browsers do when loading an html page and setting up callbacks (ping @MichaelDiBernardo).


#### blocks.js

- the purpose of units is not mentioned
- while reading the chapter, I had the webapp running at all time, trying things out. One thing that struck me is that there is a single definition of `Block` in the code, while there are clearly two very different instances of them in the application: the templates in the *Menu* area and the instances that are dragged in the *Script* area. I think the following questions should be answered:
  * how are these blocks differentiated?
  * what (in the editor) prevents us from deleting blocks in the *Menu* area

#### drag.js

- another big picture note: I think this should say that the purpose of this piece of code is to implement interactions between the *Menu* and *Script* areas, or, in other word, making a program the given language.

#### menu.js

- I find it hard to follow how this is connected to the block-based language. At this point, it feels disconnected. A definition of what is considered a script may help. E.g. a script is a sequence of operations that respects the semantics of a given language (be it block or text-based).
- I think the advice about *growing* an architecture is very valuable. Programmers often get stuck in design that tries too hard to see in the future (I'm guilty of it too :) ).  



Oh, and great to see reference to Mindstorms, more people need to know about Seymour Papert's work :) !

dethe:
@sevas I hadn't thought about connecting the blocks with ASTs or parsed languages like that. It seems obvious in retrospect. Great feedback, and a helpful outline too! Between this and @chrisseaton's feedback I think I have a good idea on how to rewrite the chapter, thanks!

And yes, more people need to know about Seymour Papert's work for sure!

MichaelDiBernardo:
Thanks to @sevas and @chrisseaton for the great feedback, and to @dethe for being so open to it. @dethe -- looking forward to seeing what you come up with :)

sevas:
@dethe @michaeldibernardo

Glad my feedback could be useful 

Looking forward to the book release!

MichaelDiBernardo:
Thank you for your changes! I think this is good to go to copy editing now.

PR #82, opened 2014-08-25
DBDB (data-store) chapter for review!

taavi:
I suggest ignoring the diff for ``chapter.md`` and just reading it directly. :)


MichaelDiBernardo:
\o/

PR #88, opened 2014-11-21
Functional database - chapter 

yoavrubin:


functionalDB/chapter.md
```diff
@@ -0,0 +1,1105 @@
...
+
+The driving force of the entity adding process is the *add-entity* function. It  is responsible to do three things - prepare the entity for addition to the database, add it to the storage and update the indices with the relevant information from the entity.
+
+Preparing an entity means providing it with an id (if it doesn’t have) and setting its timestamp field, these actions take place in the function *fix-new-entity* and it’s auxiliary functions *next-id*, *next-ts* and *update-creation-ts*. These helper functions are responsible for finding the next timestamp of the database and update the creation timestamp of a given entity (which is actually the timestamp field of each of the entity’s attributes), as well as find whether an entity needs to be assigned with an id field, if so set its id and prevent future use of that id.
+
+````clojure
+(defn- next-ts [db] (inc (:curr-time db)))
+
+(defn- update-creation-ts [ent ts-val]
+   (reduce #(assoc-in %1 [:attrs %2 :ts ] ts-val) ent (keys (:attrs ent))))
```
MichaelDiBernardo: Ah, sorry -- I misunderstood the language to mean "I'm setting the _entity's_ ts to the ts of its attributes", not "I'm setting the attribute timestamps to the current".

functionalDB/chapter.md
```diff
@@ -0,0 +1,1098 @@
...
+                                          (:storage no-ref-layer) ent))
+         new-layer (reduce (partial remove-entity-from-index ent) 
+                                 no-ent-layer (indices))]
+     (assoc db :layers (conj  (:layers db) new-layer))))
+````
+The reference removal part is handled in the *remove-back-refs* function.
+
+````clojure
+(defn- remove-back-refs [db e-id layer]
+   (let [refing-datoms (reffing-datoms-to e-id layer)
```
MichaelDiBernardo: This is a nitpick, but why do we use 1 f in the function name `refing-datoms`, and then two fs in `reffing-datoms`? :)

functionalDB/chapter.md
```diff
@@ -0,0 +1,1098 @@
...
+         remove-fn (fn[d [e a v]] (update-datom db e a v :db/remove))
+         clean-db (reduce remove-fn db refing-datoms)]
+     (last (:layers clean-db))))
+````
+It starts by locating the entities that refer to the removed entity, by calling *reffing-datoms-to* that returns a sequence of triplets, each containing the id of the referencing entity, the attribute name and the id of the removed entity.
```
MichaelDiBernardo: Why do we need the id of the removed entity? Isn't it always going to be `e-id`?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1098 @@
...
+              [reffing attr-name e-id])))
+
+````
+ When the *remove-back-refs* function gets the triplets sequence, it just calls the *update-datom* function (which will be explain in the next part) to update the referencing entity to not hold the id of the removed entity (at the found attribute).
+
+The reference removal process ends with clearing-up the removed entity’s id from the VAET index (the index that holds the references) at the *remove-entity* function.
+
+#### Updating an entity
+
+At its essence, an update is the modification of an entity’s attribute’s value. The modification itself depends on the cardinality of the attribute: an attribute whose cardinality is *:db/multiple* (meaning that the attribute holds a set of values) allows adding or removing items to/from it, or replacing the entire set with a new set. An attribute whose cardinality is *:db/single* (meaning that the attribute holds a single value) allows only to replace the value by another one.
```
MichaelDiBernardo: The first thought I had when reading this was: Why can't adding and removing elements of a set be seen as replacing the existing set with a new set which has the necessary elements added/removed?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1098 @@
...
+At its essence, an update is the modification of an entity’s attribute’s value. The modification itself depends on the cardinality of the attribute: an attribute whose cardinality is *:db/multiple* (meaning that the attribute holds a set of values) allows adding or removing items to/from it, or replacing the entire set with a new set. An attribute whose cardinality is *:db/single* (meaning that the attribute holds a single value) allows only to replace the value by another one.  
+
+Any modification of a value also needs to ripple throughout the database by modifying the various indices to reflect the change that occurred. 
+
+At this point It is important to keep in mind that the meaning of modifying here is not overwriting, but rather adding an additional layer that holds the updated database state. That updated attribute’s value in the new layer should point to the pre-update value, so we could track the changes in an entity throughout time.
+
+The update process is done at the function *update-datom*. It creates the new attribute and updates the relevant indices and ends with the creation of a new layer.
+
+````clojure
+(defn update-datom
```
MichaelDiBernardo: After reading through this a few times, I think it might be better to call this function `update-entity`, and then to explain that we use a datom as input to this function because it's how we describe the update? This would maintain naming consistency across the "low-level manipulation" API.

functionalDB/chapter.md
yoavrubin: no reason, my bad, just a typo...

functionalDB/chapter.md
yoavrubin: We don't need it, even in such a small codebase it's so hard not to have unneeded code...

functionalDB/chapter.md
yoavrubin: This is one of the best features of immutable data structures - you don't need to do a full-copy-on-modification, you pay for "in-place" modification from the caller's perspective, while all other users of that data structure (in this case - previous layers) never see any change

functionalDB/chapter.md
yoavrubin: yes, right

functionalDB/chapter.md
```diff
@@ -0,0 +1,1097 @@
...
+
+As for the function that is to be executed, this function needs to perform all the operations that the user requested, while eventually adding one layer to the database.  To do so we would need to overcome the fact that each of the lifecycle functions adds a layer reflecting the change it inflicts. 
+
+Here we resort to the insight that changes on different attributes (that can be part of either the same or different entities) accumulate when stacking layers, thus a top layer holds all the changes that built the layers below it. Therefore, the solution is to execute each of the user’s operations one after another, each creating a new layer. While doing so, we do not update the databases’s timestamp field. When the last layer is completed, we take only that top layer and place it on the initial database  (leaving all the intermediate layers to pine for the fjords) and only then update the database’s timestamp. All this is happening in the *transact-on-db* function
+
+````clojure
+(defn transact-on-db [initial-db txs]
```
MichaelDiBernardo: I think `txs` is a confusing name for this parameter. We're not really passing in transactions, are we? It's more like a list of operations that we're executing as a single transaction. Maybe `ops` or `updates`?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1097 @@
...
+       (if tx
+           (recur rst-tx (apply (first tx) transacted (rest tx)))
+           (let [initial-layer  (:layers initial-db)
+                 new-layer (last (:layers transacted))]
+             (assoc initial-db 
+                         :layers (conj  initial-layer new-layer) 
+                         :curr-time (next-ts initial-db) 
+                         :top-id (:top-id transacted))))))
+```` 
+This function is performed on the Atom that wraps the database (to modify the database state as seen by the user), but it can be used also to answer a *what-if* question - as it does not modify the given database, but rather produces a new database with the updated state. These two scenarios are supported by two execution paths that end up in the *transact-on-db* function:
```
MichaelDiBernardo: I have a suggestion for this section. Instead of discussing `what-if` and lazy argument evaluation together, I suggest:

a) Identifying that users can't really call `transact-on-db` directly, because if they provide e.g. a literal list of them in a REPL, the list will be evaluated 

b) We thus provide a higher-level `transact` call, which is actually a macro. The reason why this helps is because we basically get a form of lazy evaluation out of it. Instead of parameterizing `_transact` at this stage, I would explicitly call `*swap*` from within `_transact` to support only this use case.

c) Go through your macro expansion examples for `_transact` to describe how this works in practice.

d) Split out the `what-if` implementation into a short subsection. Here's where you can refactor `_transact` to take a post-update function.

I would also make it more explicit that `transact-on-db` requires an `*Atom*` and not a bare database (why can't `transact-on-db` call `*Atom*` on `initial-db` itself as a first step?)

I'm going to hold off on mutating this section until you've had a chance to respond to this :)

functionalDB/chapter.md
yoavrubin: Yep, ops (and op during the destructuring) is a better name, thanks !

functionalDB/chapter.md
yoavrubin: transact-on-db operates on the database - this is a bug in the writing... I will re-write this section, you are right - it missed the point, especially the fact the both transacting and what-if'ing are two very thin wrappers around the same functionality (which is what happens in transact-on-db), and are doing code reuse via higher order functions and macros.

functionalDB/chapter.md
```diff
@@ -0,0 +1,1094 @@
...
+````
+Changes into: 
+````clojure
+(_what-if my-db transact-on-db [[add-entity e3] [remove-entity e4]])
+````
+Becomes eventually: 
+````clojure
+(transact-on-db my-db  [[add-entity e3] [remove-entity e4]])
+````
+## Connected data
```
MichaelDiBernardo: This section (everything up until starting the query language), while it has some interesting parts, feels like it takes some of the momentum out of the chapter. After reading about basic CRUD operations and transactional semantics, it feels like the most exciting natural next step would be to go to the query language. 

There's quite a bit of stuff in this chapter, so I wonder how you would feel about possibly omitting this section entirely? We could then mention these two features as exercises to the reader, with the added bonus that there would be a proposed example solution in the code already :)

yoavrubin:
Do you mean the "Connected data" section? I agree that it breaks the DB-ish flow of the text, how about moving it after the query language (making it as sort of "what-else-can-we-do-here" part), and if things are still too long, follow your suggestion of just mentioning it + reader-exercise ?

MichaelDiBernardo:
Yes, I mean the "Connected Data" section. Let's go with your plan of moving it for now, and then deciding on what to do with it later.

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+Let’s start by declaring the core constructs that make up our database. 
+
+````clojure
+(defrecord Database [layers top-id curr-time])
+`````
+
+A database consists of:
+
+1. Layers of entities, each with its own unique timestamp (rings in Figure 1)
+2. A top-id value which is used to generate unique ids
```
NatalieBlack: I don't fully understand what this top-id is or how it generates ids

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+
+(defn make-entity
+   ([] (make-entity :db/no-id-yet))
+   ([id] (Entity.  id {})))
+````
+Note that if no id is given, the entity’s id is set to be *:db/no-id-yet*, which means that something else is responsible for giving it an id. We’ll see how that works later.
+
+**Attributes**
+
+Each attribute consists of its name, value, and the timestamps of both its most recent and previous update. In addition to these fields, each attribute has two fields that describe its *type* and *cardinality*.
```
NatalieBlack: so the timestamps of its two most recent updates?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+(defn add-attr [ent attr]
+   (let [attr-id (keyword (:name attr))]
+      (assoc-in ent [:attrs attr-id] attr)))
+````
+
+### Storage
+
+So far, we have talked a lot about _what_ we are going to store, without thinking about _where_ we are going to store it. In this chapter, we resort to the simplest storage mechanism, which is storing the data in memory. This is certainly not reliable, but it simplifies development and debugging and allows us to focus on more interesting parts of the program. 
+
+We will access the storage via a simple _protocol_ that will make it possible to add more durable storage providers in the future.
```
NatalieBlack: The "add more durable storage providers in the future" part threw me a bit.  Does this mean "switch to a more durable storage provider in the future", or can there be multiple storage providers at once?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+Figure 3 shows an index that would be called AVET since:
+
+* First level map holds attribute-name
+* Second level map holds the values (of the attributes)
+* Third level set holds the entity-ids (of the entities whose attribute is at the first level) 
+
+![image alt text](image_2.png)
+
+Figure 3
+
```
NatalieBlack: really liking this explanation of indexes

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+Our indexes are implemented as a map of maps, where the keys of the root map act as the first level, each such key points to a map whose keys act as the index’s second-level and the values are the index’s third level. Each element in the third level is a set, holding the leaves of the index.
+
+Each index stores the components of a datom as some permutation of its canonical 'EAV' ordering (entity_id, attribute-name, attribute-value). However, when we are working with datoms _outside_ of the index, we expect them to be in canonical format. We thus provide each index with functions *from-eav* and *to-eav* to convert to and from these orderings.
+
+In most database systems, indexes are an optional component; for example, in an RDMBS like postgresql or mysql, you will choose to add indexes only to certain columns in a table. We provide each index with a *usage-pred* function that determines whether an attribute and decides whether that attribute should be included in this index or not.
```
NatalieBlack: "that determines whether an attribute and decides whether that attribute" - this is a bit garbled, no?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+ (defn usage-pred [index] (:usage-pred (meta index)))
+````
+
+In our database there are four indexes - EAVT (as depicted in Figure 2), AVET (as can be seen in Figure 3), VEAT and VAET. We can access these as a vector of values returned from the *indexes* function.
+
+````clojure
+(defn indexes[] [:VAET :AVET :VEAT :EAVT])
+````
+
+To see how all of this comes together, the result of indexing the following five entities is visualized table below (the color coding follows the color coding of Figure 2 and Figure 3)
```
NatalieBlack: "is visualized table" -> "is visualized in the table" ?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+<span style="background-color:pink">Egypt</span> ⇒ {<span style="background-color:lightgreen">lives-in</span> ⇒ {<span style="background-color:lightblue">Cleo</span>}}</li>
+<li>
+<span style="background-color:pink">Tiber</span> ⇒ {<span style="background-color:lightgreen">river</span> ⇒ {<span style="background-color:lightblue">Rome</span>}}
+</li>
+<li>
+<span style="background-color:pink">Nile</span> ⇒ {<span style="background-color:lightgreen">river</span> ⇒ {<span style="background-color:lightblue">Egypt</span>}}
+</li></ul></td>
+  </tr>
+</table>
+Table 2
```
NatalieBlack: I appreciate these examples!

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+   (atom 
+       (Database. [(Layer.
+                   (fdb.storage.InMemory.) ; storage
+                   (make-index #(vector %3 %2 %1) #(vector %3 %2 %1) #(ref? %)); VAET                     
+                   (make-index #(vector %2 %3 %1) #(vector %3 %1 %2) always); AVET                        
+                   (make-index #(vector %3 %1 %2) #(vector %2 %3 %1) always); VEAT                       
+                   (make-index #(vector %1 %2 %3) #(vector %1 %2 %3) always); EAVT
+                  )] 0 0)))
+````
+There is one snag, though -- all collections in Clojure are immutable. Since write operations are pretty critical in a database, we call **atom** on our structure first. This is one of Clojure’s reference types, and it provides atomic writes to the element it wraps.
```
NatalieBlack: I found myself reading this paragraph several times, even though I understand what it's talking about. Something about it still tripped me up.

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+
+(defn attr-at
+   ([db ent-id attr-name] (attr-at db ent-id attr-name (:curr-time db)))
+   ([db ent-id attr-name ts] (get-in (entity-at db ts ent-id) [:attrs attr-name])))
+
+(defn value-of-at
+   ([db ent-id attr-name]  (:value (attr-at db ent-id attr-name)))
+   ([db ent-id attr-name ts] (:value (attr-at db ent-id attr-name ts))))
+
+(defn ind-at
```
NatalieBlack: is this supposed to be kind-at?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+   ([db kind] (ind-at db kind (:curr-time db)))
+   ([db kind ts] (kind ((:layers db) ts))))
+````
+
+Since we treat our database just like any other value, each of these functions take a database as an argument. Each element is retrieved by its associated identifier, and optionally the timestamp of interest. This timestamp is used in each case to find the corresponding layer that our lookup should be applied to.
+
+#### Evolution
+
+A first usage of the basic accessors is to provide a "read-into-the-past" API. This is possible as in our database, an update operation is done by appending a new layer (as oppose to overwriting). Therefore we can use the *prev-ts* property of an attribute look at the attribute at that layer, and continue going back in time looking deeper into history, thus observing the how the attribute’s value evolved throughout time.
```
NatalieBlack: "the how the" -> "how the" ?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+(_what-if my-db transact-on-db [[add-entity e3] [remove-entity e4]])
+````
+Becomes eventually: 
+````clojure
+(transact-on-db my-db  [[add-entity e3] [remove-entity e4]])
+````
+
+## Insight extraction as libraries
+
+At this point, we have the core functionality of the database in place, and it is time to add to it its raison d'être - insights extraction. The architecture approach we used here is to allow adding these capabilities as libraries, as different usages of the database would need different such mechanisms.
```
NatalieBlack: I'm not sure what exactly is meant by "insight extraction".  It seems like maybe it's just a fancy name for a simple concept?  I haven't seen it used before.

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+
+````clojure
+(defn outgoing-refs [db ts ent-id & needed-keys]
+   (let [val-filter-fn (if ref-names #(vals (select-keys ref-names %)) vals)]
+   (if-not ent-id []
+     (->> (entity-at db ts ent-id)
+          (:attrs) (val-filter-fn) (filter ref?) (mapcat :value)))))
+````
+These two functions act as the basic building blocks for any graph traversal operation, as they are the ones that raise the level of abstraction from entities and attributes to nodes and links in a graph. Once we have the ability to look at our database as a graph, we can provide various graph traversing and querying APIs. We leave this as a solved exercise to the reader - one solution can be found in the chapter's source code (see graph.clj).   
+
```
NatalieBlack: This is neat

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+   (->> items-seq ; take the items-seq
+         (map vec) ; make each collection (actually a set) into a vector
+         (reduce into []) ;reduce all the vectors into one vector
+         (frequencies) ; count for each item in how many collections (sets) it was in
+         (filter #(<= num-of-conditions (last %))) ;items that answered all conditions
+         (map first) ; take from the duos the items themselves
+         (set))) ; return it as set
+````
+
+In our example, the result of this step is a set that holds the value *1*.
```
NatalieBlack: Maybe clarify why it holds the value 1 (although I do understand why, so maybe that's not necessary)

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+
+Our journey started with a conception of a different kind of database, and ended with one that:
+
+* Supports ACI transactions (durability was lost when we decided to have the data stored in-memory) 
+* Supports "what-if" interactions
+* Answers time-related questions 
+* Handle simple datalog queries that are optimized by using indexes
+* provides APIs for graph queries
+* Introduces and implemented the notion of evolutionary queries
+
```
NatalieBlack: nice summary

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+### From archaeology to databases
+
+If we were to ask our friendly archaeologist to design a database, we might expect the requirements to reflect what would be found at an *excavation site:*
+
+* All data is found and cataloged at the site
+* Digging deeper will expose the state of things in times past 
+* Artifacts found at the same layer are from the same period
+* Each artifact will consist of state that it accumulated in different periods 
+
+For example, a wall may have Roman symbols drawn on it at on one layer, and in a lower layer there may be Greek symbols. Both these observations are recorded as part of the wall's state.
```
NatalieBlack: I'm not terribly familiar with archaeology, but at this point I'm feeling like the analogy is stretched a bit thin.  It's not taking away from my understanding at all, but I'm not sure it's adding a great deal either.

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
+# Designing a database like an archaeologist
+
+Software development is often viewed as a rigorous process, where the inputs are requirements and the output is the working product. However, software developers are people with their own perspectives and biases which color the outcome of their work. 
+
+In this chapter, we will explore how a change in a common perspective affects the design and implementation of a well-studied type of software -- a database.
```
NatalieBlack: Good opening. You have my interest :)

NatalieBlack:
Nice work!  The whole chapter was really clear (somewhat surprisingly so, considering I'm not a database person nor a clojure programmer!) without feeling like anything was over-explained.  It held my attention and I feel like I learned things. 

I left in-line comments regarding the few things that caused some confusion.

functionalDB/chapter.md
yoavrubin: I guess that this is archaeology as seen by a software guy (this is why I added that archaeology knowledge is based on Indiana Jones...). It doesn't add much, just the context that we're adding the notion of time and history to databasing

functionalDB/chapter.md
yoavrubin: How about I'll add something like "its usage is explain in section X" ?

functionalDB/chapter.md
yoavrubin: Yes, how about update the wording to say ...recent and the one before that... ?

functionalDB/chapter.md
yoavrubin: It's just an interface to plug in the storage provider. You are right, the wording here is confusing, maybe write here instead "define additional storage providers for the DB owners to select from"?

functionalDB/chapter.md
yoavrubin: Right again, too many versions ..., I'll change it to "...determines for an attribute whether it should..."

functionalDB/chapter.md
yoavrubin: Yes

functionalDB/chapter.md
yoavrubin: Actually we don't call Atom on the structure, but define it to be an Atom, which is one of Clojure's reference types. I'll update the text.

functionalDB/chapter.md
yoavrubin: No, it is short for index-at, I'll rename the function to be called indx-at

functionalDB/chapter.md
yoavrubin: Correct again, thanks :-)

functionalDB/chapter.md
yoavrubin: It's a name for all kinds of querying capabilities, as querying is how to extract insight from a database (as oppose to extract data, which is what storage systems are for)

functionalDB/chapter.md
yoavrubin: Added this: (which is the entity Id of USA), you think it would work?

MichaelDiBernardo:
Thanks for your review @NatalieBlack!

@yoavrubin -- we have one more reviewer who will be commenting also. Stay tuned :)

functionalDB/chapter.md
NatalieBlack: Sure, that would be more clear

functionalDB/chapter.md
NatalieBlack: Sounds good!

yoavrubin:
Thanks @NatalieBlack, for all the comments, they really helped making things clearer

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+
+![image alt text](image_2.png)
+
+Figure 3
+
+Our indexes are implemented as a map of maps, where the keys of the root map act as the first level, each such key points to a map whose keys act as the index’s second-level and the values are the index’s third level. Each element in the third level is a set, holding the leaves of the index.
+
+Each index stores the components of a datom as some permutation of its canonical 'EAV' ordering (entity_id, attribute-name, attribute-value). However, when we are working with datoms _outside_ of the index, we expect them to be in canonical format. We thus provide each index with functions *from-eav* and *to-eav* to convert to and from these orderings.
+
+In most database systems, indexes are an optional component; for example, in an RDMBS like postgresql or mysql, you will choose to add indexes only to certain columns in a table. We provide each index with a *usage-pred* function that determines for an attribute whether it should be included in this index or not.
```
jscn: RDMBS -> RDBMS? Might be worth defining the acronym, too, since it's the first time we've seen it. Or dropping it altogether since it's the only time it's used in the chapter.

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+ (defn usage-pred [index] (:usage-pred (meta index)))
+````
+
+In our database there are four indexes - EAVT (as depicted in Figure 2), AVET (as can be seen in Figure 3), VEAT and VAET. We can access these as a vector of values returned from the *indexes* function.
+
+````clojure
+(defn indexes[] [:VAET :AVET :VEAT :EAVT])
+````
+
+To see how all of this comes together, the result of indexing the following five entities is visualized in the table below (the color coding follows the color coding of Figure 2 and Figure 3)
```
jscn: Great! I was just about to add a comment asking for an example to make the description more concrete.

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+
+### The bare necessities
+
+The data lifecycle consists of three basic operations, which we will discuss here:
+
+* adding an entity with the *add-entity* function
+* removing an entity with the *remove-entity* function
+* updating an entity with the *update-entity* function. 
+
+Remember that, even though these functions provide the illusion of mutability, all that we are really doing in each case is adding another layer to the data. Also, since we use here Clojure's immutable data structures, we pay for such operations the price of an "in-place" change from the caller's perspective, while maintaining immutability for all other users of that data structure.
```
jscn: Can you say more about 'the price of "in-place" change from the caller's perspective while maintaing immutability...'? Why is this important? What are the consequences? I'm not familiar with Clojure so this doesn't mean much to me in its current form.

functionalDB/chapter.md
yoavrubin: Yes, correct, good catch !!

functionalDB/chapter.md
yoavrubin: I really hope it would be...

functionalDB/chapter.md
yoavrubin: change immutable to persistent data structures, this is a huge topic, so I prefer not to enter it (just say that it is the enabler for the db capabilities)

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+   (atom 
+       (Database. [(Layer.
+                   (fdb.storage.InMemory.) ; storage
+                   (make-index #(vector %3 %2 %1) #(vector %3 %2 %1) #(ref? %)); VAET                     
+                   (make-index #(vector %2 %3 %1) #(vector %3 %1 %2) always); AVET                        
+                   (make-index #(vector %3 %1 %2) #(vector %2 %3 %1) always); VEAT                       
+                   (make-index #(vector %1 %2 %3) #(vector %1 %2 %3) always); EAVT
+                  )] 0 0)))
+````
+There is one snag, though -- all collections in Clojure are immutable. Since write operations are pretty critical in a database, we define our structure to be ab **Atom**, which is is one of Clojure’s reference types, one that provides the capability of atomic writes.
```
jscn: ab **Atom** -> an **Atom**

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+````clojure
+(defn transact-on-db [initial-db ops]
+    (loop [[op & rst-ops] ops transacted initial-db]
+      (if op
+          (recur rst-ops (apply (first op) transacted (rest op)))
+          (let [initial-layer  (:layers initial-db)
+                new-layer (last (:layers transacted))]
+            (assoc initial-db :layers (conj  initial-layer new-layer) :curr-time (next-ts initial-db) :top-id (:top-id transacted))))))
+```` 
+Note here that we used the term _value_, this means that only the caller to this function is exposed to the updated state, and all other users of the database are unaware of this change (as a database is a value, and therefore cannot change).
```
jscn: There's something funky going on with the formatting in this section--for example, the numbered list on lines 549-551 are not displayed as a list at all. Do you need a blank line between lines 546 and 547?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+(defn transact-on-db [initial-db ops]
+    (loop [[op & rst-ops] ops transacted initial-db]
+      (if op
+          (recur rst-ops (apply (first op) transacted (rest op)))
+          (let [initial-layer  (:layers initial-db)
+                new-layer (last (:layers transacted))]
+            (assoc initial-db :layers (conj  initial-layer new-layer) :curr-time (next-ts initial-db) :top-id (:top-id transacted))))))
+```` 
+Note here that we used the term _value_, this means that only the caller to this function is exposed to the updated state, and all other users of the database are unaware of this change (as a database is a value, and therefore cannot change). 
+In order to have a system where users can be exposed to state changes performed by others, users do not interact directly with the database, but rather refer to it using another level of indirection [REF HERE]. This additional level is implemented using Clojure's element called *Atom*, which is a one of Clojure's reference types. Here we leverage two key features of an *Atom*, which are:
```
jscn: You mention two key features and then list three :)

I found this section hard to understand. I had to go away and read up on Clojure's *Atom* to figure it out. I think this was partly because you use the term *element* here in a different way to how you use it elsewhere in the chapter. It's also the first use of *dereference* and I get the feeling that it has a special meaning in Clojure-land. I think the 3rd point you list is the relevant one and points 1 and 2 muddy the explanation for me.

It seems that the important thing is that you pass an atom (which references the thing you want to "update") and a function (which does the updating) to the transaction, and you get back a *new* thing which is the result of the function being applied to the atom. Do I have that right?

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+<td>[:name USA #{1}]</td><td>["?e" nil "?nm"]</td>
+</tr>
+<tr>
+<td>[:birthday "July 4, 1776" #{1}]</td><td>["?e" nil "?bd"]</td>
+</tr>
+<tr>
+<td>[:speak "English" #{1}]</td><td>["?e" nil nil]</td>
+</tr>
+</table>
+We are now ready to being reporting the results. The result clause structure is unwieldy for this purposes, so we will convert it into an an index-like structure (map of maps) -- with a significant twist.
```
jscn: unwieldy for this purposes -> unwieldy for *these* purposes **OR** unwieldy for this *purpose*

functionalDB/chapter.md
```diff
@@ -0,0 +1,1138 @@
...
+</table>
+
+This structure acts as the query that is executed in a later phase, once the engine decides on the right plan of execution.
+
+#### Phase 2 - Making a plan
+In this phase, we inspect the query to determine a good plan that will produce the result it describes.
+
+In general, this will involve choosing the appropriate index and constructing a plan in the form of a function.
+
+We choose the index based on the _single_ joining variable (that can operate on only one kind of elements)
```
jscn: one kind of elements -> one kind of element

jscn:
I found myself a bit lost as the chapter went along I think because I couldn't really understand the later code samples. I've never read any Clojure before and while I could figure out the early snippets, I couldn't understand the later ones. I'm afraid I don't have a good suggestion for how to improve that, but thought I should note is as a concern.

It might be nice to see a worked example of updating a database's content over time, and a query which gives different results when different times are specified, since that was one of the main rationales for the database. I can understand that it's possible by reading the chapter and thinking about it, but it would be nice to see a concrete example to prove it :)

All in all, a very interesting premise, the prose reads well and the explanations are clear, despite some of the code being hard to understand (for me at least).

functionalDB/chapter.md
MichaelDiBernardo: Yes it will -- we'll figure it out :)

functionalDB/chapter.md
MichaelDiBernardo: We'll be rendering these documents after changing the formatting a bit on our end anyhow, so we can fix it then!

MichaelDiBernardo:
Thank you @jscn!

functionalDB/chapter.md
yoavrubin: I updated the explanation (and the count of the features...), I think / hope it is clearer now. The point here  is to different between an atom, which can change in a managed way (transaction) to a value (that cannot be changed).

functionalDB/chapter.md
yoavrubin: yes, thanks for that catch

functionalDB/chapter.md
yoavrubin: Changed it to single kind of elements (as there may be many elements of the same kind)

yoavrubin:
Thanks @jscn for the review. 
I agree that most of the snippets are hard to read for people who are not familiar with Clojure. This is something that is common for almost any Clojure snippet, and all Clojurians are just waiting for the rest of the world to re-align its thinking and to start using this language. I hope that I did my part by showing some of its powers. 
As for example, there is one in the chapter's repo that I used for testing, however, as I didn't count it as part of the 500 lines limit, I also didn't explain on it in the writing. I'm sure that anyone who want to further explore chapter and its the code base will find it.

MichaelDiBernardo:
I think this is ready for copy-edit now -- thanks to the reviewers for the quick turnaround on this :)

@yoavrubin: There are two stages to this process. The first involves our copy editor @amyrbrown  rewording / cleaning up paragraphs (much like what we did together last month), and the second involves formatting and typesetting it. We will open a PR here at each stage and notify you so that you get a chance to comment on our changes.

PR #95, opened 2015-01-07
Update web crawler example code.

ajdavis:
Use "aiohttp" package instead of reimplementing HTTP protocol. This lets us simplify the example to focus on async I/O and coroutines.

Refactor the control flow to use a JoinableQueue since that's my specialty.

### DIFF
--- a/crawler/crawling.py
+++ b/crawler/crawling.py
@@ -211,17 +211,17 @@ class Crawler:
             return False
         parts = urllib.parse.urlparse(url)
         if parts.scheme not in ('http', 'https'):
-            logger.info('skipping non-http scheme in %r', url)
+            logger.debug('skipping non-http scheme in %r', url)
             return False
         host, port = urllib.parse.splitport(parts.netloc)
         if not self.host_okay(host):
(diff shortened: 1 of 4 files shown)


MichaelDiBernardo:
This looks good to me. I especially like that the unit of work being placed on the queue is now a URL rather than a connection. 

Given that your changes are incremental and that the general context around your changes has already been heavily reviewed, I'm good to merge this now and have other tech reviewers look at the code changes + the chapter text at the same time.

crawler/crawling.py
```diff
@@ -1,14 +1,16 @@
...
 import cgi
-from http.client import BadStatusLine
+from collections import namedtuple
 import logging
 import re
 import time
 import urllib.parse
 
-logger = logging.getLogger(__name__)
+import aiohttp  # Install with "pip install aiohttp".
```
MichaelDiBernardo: I think it's safe to add a `requirements.txt` to the project, and to put a quick note in the readme on how to `pip install -r`.

ajdavis: Good idea.

PR #105, opened 2015-02-11
Sampling edits

amyrbrown:


sampler/README.md
```diff
@@ -128,18 +130,18 @@ points distributed across different stats (e.g., +2 wisdom and +3
 intelligence) or concentrated within a single stat (e.g., +5
 charisma).
 
-How would we randomly sample these stats? The easiest way is probably
-to first sample the overall item bonus. Then, we sample the way the
+How would we randomly sample this distribution? The easiest way is probably
```
amyrbrown: Please let me know if this changes your meaning.

sampler/README.md
```diff
@@ -437,17 +439,17 @@ with our sampling function, then they should approximate the exact PDF
 or PMF. If after many samples the approximation is poor or obviously
 wrong, then we know there is a bug in our code somewhere.
 
-Another reason to implement the PMF or PDF is because frequently, you
+Another reason to implement the PMF or PDF is that frequently, you
```
amyrbrown: "the reason is because" isn't strictly incorrect, but it's inelegant enough to change in this context. (It would be fine in a less formal context.)

sampler/README.md
```diff
@@ -471,7 +473,7 @@ $$
 p(\mathbf{x}; \mathbf{p}) = \frac{\Gamma((\sum_{i=1}^k x_i)+1)}{\Gamma(x_1+1)\cdots{}\Gamma(x_k+1)}p_1^{x_1}\cdots{}p_k^{x_k},
 $$
 
-#### Working in "log-space"
+#### Working in "Log-Space"
```
amyrbrown: I'm not sure of your use of "log-space" here. You seem to be using it to mean "working with logs", but when I looked it up, all the references seemed to be to " complexity class containing decision problems ... logarithmic amount of memory space" (Wikipedia). Will your usage be familiar to our audience? (Although even if it isn't, I think you have explained it clearly enough at the start of the section.)

sampler/README.md
```diff
@@ -577,7 +579,7 @@ def log_pmf(self, x):
     The evaluated log-PMF for draw `x`
 
     """
-    # Get the total number of events.
+    # Get the total number of events
```
amyrbrown: I removed terminating periods from short comments for consistency.

sampler/README.md
jhamrick: I would say "sample from" rather than just "sample" but otherwise this change is fine.

sampler/README.md
jhamrick: Hmm, you're right, though I haven't seen any alternate terms for this concept. Maybe "working with log values" would be clearer and avoid confusion with the complexity class?

jhamrick:
These changes look great, thanks! The only other thing would be that any changes to the code in the text should also be made to the code files. I guess I don't know how the final format of the book is going to be, but I was imagining those files would be available for people to download and run, in which case they should be the same as the code in the text.

amyrbrown:
Good point about the code. I'll match the code to the changes in the chapter and add it to this PR.

sampler/README.md
amyrbrown: I've taken out references to "log-space" and done some minor rewriting to accommodate them. (Sometimes tricky -- the metaphor of being "in" log-space is useful!) Can you look over this section to check that it reads smoothly and I haven't mangled your meaning?

jhamrick: Looks great to me!

jhamrick:
Cool, I just looked over everything and it sounds good to me!

amyrbrown:
Awesome. @MichaelDiBernardo, this one is ready to merge!

amyrbrown:
(Dammit, hit "close and comment" by mistake.)

MichaelDiBernardo:
Thanks to you both :)

PR #112, opened 2015-03-06
Objmodel edits

amyrbrown:
Copyedits for the Object Model chapter -- please see comments on PR for more details and queries.


objmodel/chapter.md
```diff
@@ -1,86 +1,90 @@
-
+<!-- English spelling: "behaviour" line 46 -->
 A Simple Object Model
 ======================
 
 
-Introduction
-----------------
+## Introduction
```
amyrbrown: I changed all the hyphen-style header markings to hashes because Git's markdown doesn't accept "++++" as a second-level header. I thought it better to use the same system throughout.

objmodel/chapter.md
```diff
@@ -1,86 +1,90 @@
...
-interesting. It often reveals the family resemblance between different
-languages. It can be useful to quickly understand the model of a new language by
-putting it into the context of the models of other languages and it can give a
-better feeling of the programming language design space.
+interesting. They often reveal the family resemblance between different
+languages. 
+It can be useful to put the model of a new language into 
+the context of the models of other languages, 
+both to quickly understand the new model, and to get a
+better feeling for the programming language design space.
```
amyrbrown: Sometimes I rearranged your sentences for clarity and conciseness -- please read my changes carefully and let me know if I have changed your meaning by mistake!

objmodel/chapter.md
```diff
@@ -1,86 +1,90 @@
...
 object model.
 
-The object model of this section will have classes and instances of them, the
+The object model in this section will have classes and instances of them, the
 ability to read
-and write attributes into objects, the ability to call methods on objects and
+and write attributes into objects, the ability to call methods on objects, and
 the ability for a class to be a subclass of another class. Right from the
-beginning, classes will be completely regular objects, that can themselves have
+beginning, classes will be regular objects that can themselves have
```
amyrbrown: I removed "completely" here because "regular" doesn't gain much from being modified by "completely". If "completely regular" has some specific meaning in Python (apart from the non-technical sense of "absolutely normal" then we should leave "completely" in.

objmodel/chapter.md
```diff
@@ -1,86 +1,86 @@
...
-A note on terminology: In this chapter I will use the word "instance" to mean
-"an object that is not a class".
 
-To start the implementation a good approach is to write a test to think about
+A good approach to start with is to write a test to specify
 what the to-be-implemented behaviour should be. All tests presented in this
-chapter will always consist of two parts: A bit of completely regular Python
-code defining and using a few classes, making use of increasingly advanced
-features of the Python object model. The second half of each test is the
+chapter will consist of two parts. First, a bit of regular Python
```
amyrbrown: When you introduce something as having a number of parts, it's helpful to the reader if you are very clear about what those parts are as you come to them.

objmodel/chapter.md
```diff
@@ -1,86 +1,86 @@
...
-be done manually in the tests. E.g. instead of writing ``obj.attribute`` in
+be done manually in the tests. For example, instead of writing ``obj.attribute`` in
 Python, in the object model we would use a method ``obj.read_attr("attribute")``.
-This mapping would in a real language implementation be done
-by the interpreter of the language, or a compiler. A further simplification in
-this chapter will be that we make no sharp distinction between the code that
+This mapping would, in a real language implementation, be done
+by the interpreter of the language, or a compiler. 
+
+A further simplification in
```
amyrbrown: Generally it's good to break up paragraphs as often as you can, especially in technical texts like this. It makes things easier for the reader.

objmodel/chapter.md
```diff
@@ -119,10 +119,10 @@ def test_read_write_field():
     assert obj.read_attr("b") == 5
 ````
 
-The test uses three things that we will have to implement in this chapter. On
-the one hand the classes ``Class`` and ``Instance`` that represent classes and
-instances of our object model, respectively. On the other hand, there are two
-special instances of class, ``OBJECT`` and ``TYPE``. ``OBJECT`` corresponds to
+The test uses three things that we have to implement. 
+The classes ``Class`` and ``Instance`` represent classes and
```
amyrbrown: I rewrote this to remove "on the one hand" and "on the other hand" because you start off saying there are three things, and I didn't want our readers to be distracted by the thought of somebody with three hands.

objmodel/chapter.md
```diff
@@ -232,15 +232,14 @@ TYPE.cls = TYPE
...
 ````
 
-To define new metaclasses it is enough to subclass ``TYPE``. However, in the
-rest of this chapter we won't do that, and simply always use ``TYPE`` as the
+To define new metaclasses, it is enough to subclass ``TYPE``. However, in the
+rest of this chapter we won't do that; we'll simply always use ``TYPE`` as the
 metaclass of every class.
 
-XXX diagram: figures/inheritance.svg
+![](figures/inheritance.png)
```
amyrbrown: I converted the SVG files into PNGs just so I could check flow. @MichaelDiBernardo, we will probably want to do a cleaner conversion later.

objmodel/chapter.md
```diff
@@ -885,44 +878,44 @@ class Instance(Base):
             self.map = new_map
 ````
 
-The class now passes ``None`` as the fields dict to ``Base``, as ``Instance``
-will now store the content of the dict in another way. Therefore it now needs
-to override the the ``_read_dict`` and ``_write_dict`` methods. In a real
-implementation, we would refactor the ``Base`` class to not be responsible for
-storing the field dictionary any more, but for this section having instances
+The class now passes ``None`` as the fields dictionary to ``Base``, as ``Instance``
+will store the content of the dictionary in another way. Therefore it needs
```
amyrbrown: I'm not sure if you are using "dict" as a short form for "dictionary" here, or if it has special meaning in this context. I assumed the former; if the latter, please let me know.

objmodel/chapter.md
```diff
@@ -1,86 +1,86 @@
-
+<!-- English spelling: "behaviour" line 46 -->
```
cfbolz: I was confused by that comment. is that about British vs. American spelling? I'm usually trying to do American (see all the -ize), particularly since British requires "object-orientated" which is just weird :-).

objmodel/chapter.md
cfbolz: Both. The Python type for dictionary is called "dict". Should we be consistent about expanding it to "dictionary" everywhere? I also wonder whether we should add a parenthese at the first use, because the same data type is called "hashmap" in many other languages.

objmodel/chapter.md
```diff
@@ -1,86 +1,86 @@
...
+The object model in this section will have classes and instances of them, the
 ability to read
-and write attributes into objects, the ability to call methods on objects and
+and write attributes into objects, the ability to call methods on objects, and
 the ability for a class to be a subclass of another class. Right from the
-beginning, classes will be completely regular objects, that can themselves have
+beginning, classes will be regular objects that can themselves have
 attributes and methods.
-A note on terminology: In this chapter I will use the word "instance" to mean
-"an object that is not a class".
```
cfbolz: I know that this note is a bit clunky. But I don't think we should remove it completely, it is the result of independent requests by several of the technical reviewers.

The reason why this is not completely obvious is that "classes are just regular objects", so a class might well be called an instance as well, of its metaclass.

objmodel/chapter.md
```diff
@@ -202,26 +202,26 @@ class Class(Base):
...
+object, they (indirectly) inherit from ``Base``. Thus, the class needs to be an instance of
+another class: its metaclass.
 
-Now our first test above almost passes. The only missing bit is the definition
+Now our first test almost passes. The only missing bit is the definition
 of the base classes ``TYPE`` and ``OBJECT``, which are both instances of
 ``Class``. For these we will make a major departure from the Smalltalk model,
-which has a fairly complex metaclass system. Instead we will use the model of ObjVlisp, which is
-also the one that Python adopted (footnote: P. Cointe, “Metaclasses are first
+which has a fairly complex metaclass system. Instead we will use the model used by ObjVlisp, which
```
cfbolz: Maybe "Instead we will use the model introduced by ObjVlisp"? I dislike the double "use".

objmodel/chapter.md
cfbolz: Yes, the current conversion somehow lost the arrowheads on the lines, so we must make better ones later.

objmodel/chapter.md
```diff
@@ -885,44 +877,44 @@ class Instance(Base):
...
+The class now passes ``None`` as the fields dictionary to ``Base``, as ``Instance``
+will store the content of the dictionary in another way. Therefore it needs
+to override the ``_read_dict`` and ``_write_dict`` methods. In a real
+implementation, we would refactor the ``Base`` class so that it is no longer responsible for
+storing the fields dictionary, but for now having instances
 store ``None`` there is good enough.
 
-When creating a new instance it starts out with using the ``EMPTY_MAP``, which
-has no attributes, and an empty storage. To implement ``_read_dict``, the
+When creating a new instance, the class starts out using the ``EMPTY_MAP``, which
```
cfbolz: This is wrong. It's really the *instance* that starts out using the ``EMPTY_MAP``.

objmodel/chapter.md
```diff
@@ -885,44 +877,44 @@ class Instance(Base):
...
+attributes will have a larger memory footprint than if we just use dictionaries.
 
-This is a common approach for optimizing dynamic languages. It is often not
-possible to find optimizations that are faster or use less memory in *all*
-cases. To still speed up programs in practice, optimizations are chosen that
-should apply to how the language is typically used in practice, while potentially
-making the behaviour worse for programs that use extremely dynamic features.
+This is a common problem when optimizing dynamic languages. It is often not
+possible to find optimizations that are faster or use less memory in all
+cases. In practice, optimizations are made that
```
cfbolz: I think I prefer "chosen" instead of "made" here, to really make clear that it's a conscious choice by the language implementer, who picks optimizations to use out of a (usually infinite) set of possible ones.

objmodel/chapter.md
cfbolz: Ok, I indeed mean "absolutely normal". I still think I would prefer some term that is a bit stronger than just "regular": the idea that classes are themselves just regular objects is a rather radical one, and historically not obvious.

objmodel/chapter.md
MichaelDiBernardo: If we're referring explicitly to the python type (and not the general concept of the dictionary as a data-structure), I think `dict` is just fine.

objmodel/chapter.md
MichaelDiBernardo: I also think this is important to keep in, even if the wording needs to change.

objmodel/chapter.md
amyrbrown: Ah, I see. Looking at the definition of regular, I realize that it actually has several subtly different meanings. I suggest we use "ordinary", and change it to:

"...classes will be completely ordinary objects that can..."

I think "ordinary" captures the idea that classes aren't a special thing, and "completely" acknowledges that you realize this is a surprising assertion, and yes, you really mean it.

What do you think?

objmodel/chapter.md
amyrbrown: I think in most cases here we should expand to "dictionary", when we're referring to the fields dictionary data structure. I agree with Mike that if we're talking about Python's `dict` the abbreviation is appropriate, but most of the time here we're referring to a dictionary implemented using `dict`.

objmodel/chapter.md
amyrbrown: Yes, that's a note to myself about spelling. I'm afraid you let quite a few stray "u"s slip in in "behaviour", and -ize and object-oriented are quite acceptable in British English. The OED is quite forgiving. :)

objmodel/chapter.md
amyrbrown: Agreed -- I removed it by mistake -- I think I must have meant to move it, and then forgot to paste it after cutting. Thanks for catching it.

objmodel/chapter.md
amyrbrown: Agreed.

objmodel/chapter.md
amyrbrown: Okay, how about "A newly created instance starts out using the..."

objmodel/chapter.md
amyrbrown: Fair enough.

amyrbrown:
I will make the changes above and update the PR -- apart from the changes around line 64 and 887, which I'm waiting for @cfbolz's response about.

objmodel/chapter.md
cfbolz: "...classes will be completely ordinary objects that can..." sounds good to me.

objmodel/chapter.md
cfbolz: yes, excellent.

cfbolz:
Ok, the updated pull request looks good. if @amyrbrown changes the last two things, this looks great to me.

amyrbrown:
I made those last two changes (sorry for the delay) -- so this is ready to merge, @MichaelDiBernardo. Thanks @cfbolz!

cfbolz:
Thanks @amyrbrown!

MichaelDiBernardo:
Thanks very much to you both :)

PR #116, opened 2015-04-02
Copyedits to the Paxos clustering chapter

amyrbrown:
See comments and queries inline. Some of my changes are converting from RST to MD (which is not complete), there are some spelling fixes, and other things.

### DIFF
--- a/cluster/CHAPTER.rst
+++ b/cluster/CHAPTER.rst
@@ -1,20 +1,20 @@
-Clustering by Consensus
-***********************
+<!-- American spelling -->
+# Clustering by Consensus
+
 
 In this chapter, we'll explore implementation of a network protocol designed to support reliable distributed computation.
 Network protocols can be difficult to implement correctly, so we'll look at some techniques for minimizing bugs and for catching and fixing the remaining few.
 Building reliable software, too, requires some special development and debugging techniques.
 
-Motivating Example
-==================
+## Motivating Example
 
 The focus of this chapter is on the protocol implementation, but as a motivating example let's consider a simple bank account management service.
 In this service, each account has a current balance and is identified with an account number.
 Users access the accounts by requesting operations like "deposit", "transfer", or "get-balance".
 The "transfer" operation operates on two accounts at once -- the source and destination accounts -- and must be rejected if the source account's balance is too low.
 
-If the service is hosted on a single server, this is easy to implement: use a lock to make sure that transfer operations don't run in parallel, and verify the souce account's balance in that method.
-However, a bank cannot rely on a single server for its critical account balances!
+If the service is hosted on a single server, this is easy to implement: use a lock to make sure that transfer operations don't run in parallel, and verify the source account's balance in that method.
+However, a bank cannot rely on a single server for its critical account balances.
 Instead, the service is *distributed* over multiple servers, with each running a separate instance of exactly the same code.
 Users can then contact any server to perform an operation.
 
@@ -24,11 +24,10 @@ But this approach introduces a serious failure mode: if two servers process oper
 Even if the servers share operations with one another instead of balances, two simultaneous transfers out of an account might overdraw the account.
 
 Fundamentally, these failures occur when servers use their local state to perform operations, without first ensuring that the local state matches the state on other servers.
-For example, imagine that server A receives a transfer operation from account 101 to account 202, when server B has already processed another transfer account 101's full balance to account 202, but not yet informed server A.
+For example, imagine that server A receives a transfer operation from account 101 to account 202, when server B has already processed another transfer of account 101's full balance to account 202, but not yet informed server A.
 The local state on server A is different from that on server B, so server A incorrectly allows the transfer to complete, even though the result is an overdraft on account 101.
 
-Distributed State Machines
-==========================
+## Distributed State Machines
 
 The technique for avoiding such problems is called a "distributed state machine".
 The idea is that each server executes exactly the same deterministic state machine on exactly the same inputs.
@@ -37,7 +36,7 @@ Operations such as "transfer" or "get-balance", together with their parameters (
 
 The state machine for this application is simple:
 
-.. code-block:: python
+```python
 
     def execute_operation(state, operation):
         if operation.name == 'deposit':
@@ -53,9 +52,10 @@ The state machine for this application is simple:
             return state, True
         elif operation.name == 'get-balance':
             return state, state.accounts[operation.account]
+```
 
 Note that executing the "get-balance" operation does not modify the state, but is still implemented as a state transition.
-This guarantees that the returned balance is the latest information in the cluster of servers, not based on the (possibly stale) local state on a single server.
+This guarantees that the returned balance is the latest information in the cluster of servers, and is not based on the (possibly stale) local state on a single server.
 
 This may look different than the typical state machine you'd learn about in a computer science course.
 Rather than a finite set of named states with labeled transitions, this machine's state is the collection of account balances, so there are infinite possible states.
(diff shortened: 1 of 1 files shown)


cluster/CHAPTER.rst
```diff
@@ -24,11 +24,10 @@ But this approach introduces a serious failure mode: if two servers process oper
 Even if the servers share operations with one another instead of balances, two simultaneous transfers out of an account might overdraw the account.
 
 Fundamentally, these failures occur when servers use their local state to perform operations, without first ensuring that the local state matches the state on other servers.
-For example, imagine that server A receives a transfer operation from account 101 to account 202, when server B has already processed another transfer account 101's full balance to account 202, but not yet informed server A.
-The local state on server A is different from that on server B, so server A incorrectly allows the transfer to complete, even though the result is an overdraft on account 101.
+For example, imagine that server A receives a transfer operation from Account 101 to Account 202, when server B has already processed another transfer of Account 101's full balance to Account 202, but not yet informed server A.
```
amyrbrown: I'm not sure that this example is necessary -- I think you've already made the problem clear at this point. If length is a concern, you could cut this example.

cluster/CHAPTER.rst
```diff
@@ -170,8 +170,7 @@ In concrete terms, this means adding a slot number to each message, with all of
 Running Paxos for every slot, with its minimum of two round trips, would be too slow.
 Multi-Paxos optimizes by using the same set of ballot numbers for all slots, and performing the ``Prepare``/``Promise`` phase for all slots at once.
 
-Paxos Made .. Pretty Hard, Actually
------------------------------------
+### Paxos Made… Pretty Hard, Actually
```
amyrbrown: @MichaelDiBernardo I put an ellipsis in here -- I'm not sure how that will be encoded in LaTeX or epub, so let's keep an eye on it.

cluster/CHAPTER.rst
```diff
@@ -200,24 +199,22 @@ Furthermore, we need some way to initialize a new cluster.
...
 Let's get started.
 
-Types and Constants
--------------------
+### Types and Constants
 
-Cluster's protocol uses 15 different message types, each defined as a Python ``namedtuple``.
+Cluster's protocol uses fifteen different message types, each defined as a Python ``namedtuple``.
 
 {{{ from_to cluster.py '# message types' '^$'
```
amyrbrown: @MichaelDiBernardo I left this RST in because it looks useful and I'm not sure how to preserve the information in Markdown (or even if we want/need to preserve the information). Can I leave it for you?

cluster/CHAPTER.rst
```diff
@@ -846,18 +835,16 @@ It responds to the leader with either ``Decided`` or ``Preempted``.
     
 }}}
 
-.. note::
 
-    A surprisingly subtle bug appeared here during development.
```
amyrbrown: We're not including sidebars in this book, so I removed the indentations here and added some text ("As an aside") to allow it to flow as a regular paragraph.

cluster/CHAPTER.rst
```diff
@@ -1269,11 +1248,9 @@ When something *does* break, all of the debugging support becomes critical: if t
 
 I performed some manual fuzz testing of cluster during development, but a full fuzz-testing infrastructure is beyond the scope of this project.
 
-Implementation Challenges
-=========================
+## Implementation Challenges
 
-Follow the Leader
```
amyrbrown: The "Implementation Challenges" section only has one subsection, "Follow the Leader" -- would it be possible to move the Follow the Leader section up into an earlier part of the chapter? Is there somewhere else it would make sense?

cluster/CHAPTER.rst
djmitche: Are you looking at the {{{ specifically?  If so, that can get stripped -- it's part of my system for including the actual code in the chapter, and the code isn't changing anymore.

cluster/CHAPTER.rst
djmitche: IIRC, the content reviewer asked me to add an example here.  Left to my own devices I tend to err on the side of too little, so I'd prefer to leave it in.  But you're right, if space is a priority, it's a good first target.

cluster/CHAPTER.rst
djmitche: Hah, good catch.  I think I had other challenges in mind (perhaps never written down). Let's leave it here, and just rename it to something colorful like "Power Struggles".  I think it fits in nicely between the more mundane testing stuff and the as-yet-unsolved "Further Extensions" section.

cluster/CHAPTER.rst
```diff
@@ -63,24 +63,22 @@ Still, the usual rules of deterministic state machines apply: starting with the
...
+Multi-Paxos builds on this foundation by agreeing on a numbered sequence of facts, one at a time.
+To implement a distributed state machine, we use Multi-Paxos to agree on each state-machine input, and execute them in sequence.
 
-Simple Paxos
-------------
+### Simple Paxos
 
 So let's start with "Simple Paxos", also known as the Synod protocol, which provides a way to agree on a single value that can never change.
-The name Paxos comes from the mythical island in "The Part-Time Parliament", where lawmakers vote on legislation through a process Lamport dubbed a Synod.
+The name Paxos comes from the mythical island in "The Part-Time Parliament", where lawmakers vote on legislation through a process Lamport dubbed the Synod protocol.
```
djmitche: I think "Synod" in Lamport's language has the same sense as "Hearing", "Roll-Call Vote", or "Baptism".  A bunch of lawmakers show up and hold a Synod.  To my knowledge there's nothing called "the Synod protocol".

cluster/CHAPTER.rst
```diff
@@ -244,11 +241,11 @@ Using named tuples to describe each message type keeps the code clean and helps
 The named tuple constructor will raise an exception if it is not given exactly the right attributes, making typos obvious.
 The tuples format themselves nicely in log messages, and as an added bonus don't use as much memory as a dictionary.
 
-Creating a message reads naturally::
+Creating a message reads naturally:
```
djmitche: The double-colon implicitly formats the following indented block as Python, so the below should probably be flagged with \`\`\`python...```

cluster/CHAPTER.rst
```diff
@@ -279,12 +276,11 @@ Finally, Cluster uses two named data types, named to correspond to the protocol
...
 }}}
 
-Component Model
----------------
+### Component Model
 
 Humans are limited by what we can hold in our active memory.
-We can't reason about the entire Cluster implementation at once -- it's just too much, and too easy to miss details.
-For similar reasons, large monolithic codebases are harder to test: test cases must manipulate many moving pieces and are brittle, failing on almost any change to the code.
+We can't reason about the entire Cluster implementation at once -- it's just too much, so it's too easy to miss details.
```
djmitche: Probably remove the second `too` here?

djmitche:
This is great -- thank you so much!

cluster/CHAPTER.rst
MichaelDiBernardo: Alternately, I think 'Paxos Made Pretty Hard' could also work?

cluster/CHAPTER.rst
MichaelDiBernardo: For the record, I did think about stealing this :)

cluster/CHAPTER.rst
djmitche: :+1: to that suggestion

cluster/CHAPTER.rst
amyrbrown: Okay, let's leave it for now.

cluster/CHAPTER.rst
amyrbrown: In the original paper "The Part-Time Parliament", Lamport uses Synod the way it's defined in the dictionary, "an assembly of the clergy":

> The Paxon Parliament evolved from an earlier ceremonial Synod of priests...

He refers to the procedure they use to make decisions as "the Synod protocol":

>The resulting Synod protocol is described here ... Mathematicians derived the Synod protocol in a series of steps ... The complete Synod protocol, satisfying the consistency and progress requirements, was obtained by restricting the basic protocol. ... etc.

However, the Lamport paper also refers to a Parliamentary protocol (Section 3), which is derived from the Synod protocol. Determining whether your algorithm implements the Synod protocol or the Parliamentary protocol is beyond the scope of this copyedit. :)

cluster/CHAPTER.rst
amyrbrown: Cool, I will push that change.

cluster/CHAPTER.rst
amyrbrown: Cool, thanks -- I was stripping the RST without really knowing what I was doing. >_<

cluster/CHAPTER.rst
amyrbrown: Good call.

cluster/CHAPTER.rst
amyrbrown: I've cleaned out all those {{{ -- very clever!

cluster/CHAPTER.rst
djmitche: I stand very much corrected - thanks :)

The Parliamentary protocol in section 3 is multi-decree, so very similar to MultiPaxos -- which means the Synod protocol is the correct term for this, then "Simple Paxos" section.

cluster/CHAPTER.rst
MichaelDiBernardo: We're going to have to re-typeset most code anyways :)

MichaelDiBernardo:
This looks great. @djmitche do you have any outstanding comments, or are you OK with merging?

djmitche:
SHIP IT :)

MichaelDiBernardo:
Amazing :)

PR #117, opened 2015-04-03
Fix an incorrect end tag, thanks NPrescott

nedbat:
Oops, let's fix this mistake.  `{% for %}` has to be paired with `{% endfor %}`


MichaelDiBernardo:
Yes, thanks @NPrescott!

PR #124, opened 2015-05-04
Edits for the functional DB chapter

amyrbrown:
Lots of changes here, but many are formatting or minor punctuation changes. Also some clarification and streamlining of language.


functionalDB/chapter.md
```diff
@@ -1,103 +1,104 @@
...
 
 ## Introduction 
 
-Database systems are designed to store and query data. This is something that all information workers do; however, the systems themselves were designed by computer scientists. As a result, modern database systems are highly influenced by what a computer scientist’s definition of data is, and what can be done with it. 
+Database systems are designed to store and query data. This is something that all information workers do; however, the systems themselves were designed by computer scientists. As a result, modern database systems are highly influenced by computer scientists’ definition of what data is, and what can be done with it. 
 
-For example, most modern databases implement updates by overwriting old data in-place instead of appending the new data and keeping the old. This mechanism, nicknamed "place oriented programming" by Rich Hickey [REF], saves storage space but makes it impossible to retrieve the entire history of a particular record. This design decision reflects the computer scientist’s perspective that ‘history’ is less important than the price of its storage.
```
amyrbrown: We still need this reference.

functionalDB/chapter.md
```diff
@@ -1,103 +1,104 @@
...
-* The entire circle is the excavation site
-* Each ring is a _layer_ (here numbered from 0 to 4) 
-* Each slice is a labeled artifact (‘a’ through ‘e’)
-* Each artifact has a ‘color’ attribute (where white means that no update was made)
-* Black arrows denote a change in color between layers (e.g., from c.color @t2 to c.color @t0)
-* Light blue arrows are arbitrary relationships of interest between entities (e.g., from ‘b’ to ‘d’)
+* The entire circle is the excavation site.
+* Each ring is a _layer_ (here numbered from 0 to 4).
+* Each slice is a labeled artifact (‘a’ through ‘e’).
+* Each artifact has a ‘color’ attribute (where white means that no update was made).
```
amyrbrown: The book will be printed in black and white (greyscale) -- so these figures and the relevant text need to be changed.

functionalDB/chapter.md
```diff
@@ -108,27 +109,28 @@ Similar to entities, creating an attribute is done using the *make-attr* functio
      {:pre [(contains? #{:db/single :db/multiple} cardinality)]}
     (with-meta (Attr. name value -1 -1) {:type type :cardinality cardinality})))
 ````
+
 There are a couple of interesting patterns used in this constructor function: 
 
-* We use Clojure’s _Design by Contract_ [ADD REF HERE] pattern to validate that the cardinality parameter is a permissible value
-* We use Clojure’s destructuring mechanism to provide a default value of *:db/single* if one is not given
-* We use Clojure’s metadata capabilities to separate between an attribute's data (name, value and timestamps) and its metadata (type and cardinality). In Clojure, metadata handling is done using the *with-meta* (setting metadata) and *meta* (reading metadata) functions.
+* We use Clojure’s _Design by Contract_ [ADD REF HERE] pattern to validate that the cardinality parameter is a permissible value.
```
amyrbrown: Still need this reference.

functionalDB/chapter.md
```diff
@@ -146,40 +148,37 @@ And here's our in-memory implementation of the protocol, which uses a map as the
...
+If you've used a database system before, you are probably already familiar with the concept of an _index_, which is a supporting data structure that consumes extra space in order to decrease the average query time.  In our database, an index is a three-leveled structure which stores the components of a datom in a specific order. Each index derives its name from the order it stores the datom's components in.
 
 For example, let’s look at at the index sketched in Figure 2:
-* the first level stores entity-ids (the blue-ish area) 
-* the second level stores the related attribute-names (the green-ish area)
-* the third level stores the related value (the pink-ish area)
 
-This index is named EAVT, as the top level map holds (E) entity ids, the second level holds (A) attribute names, and the leaves hold (V) values. The (T) comes from the fact that each layer in the database has its own indexes, hence the index itself is relevant for a specific (T) time. 
+* The first level stores entity-IDs (the blueish area).
```
amyrbrown: Again, we can't use colours for the print book or epubs.

functionalDB/chapter.md
```diff
@@ -190,13 +189,14 @@ In most database systems, indexes are an optional component; for example, in an
...
-In our database there are four indexes - EAVT (as depicted in Figure 2), AVET (as can be seen in Figure 3), VEAT and VAET. We can access these as a vector of values returned from the *indexes* function.
+In our database there are four indexes: EAVT (see Figure 2), AVET (see Figure 3), VEAT and VAET. We can access these as a vector of values returned from the `indexes` function.
 
 ````clojure
 (defn indexes[] [:VAET :AVET :VEAT :EAVT])
 ````
 
-To see how all of this comes together, the result of indexing the following five entities is visualized in the table below (the color coding follows the color coding of Figure 2 and Figure 3)
+<!-- FIXME: more colours -->
+To demonstrate how all of this comes together, the result of indexing the following five entities is visualized in the table below (the color coding follows the color coding of Figure 2 and Figure 3).
```
amyrbrown: One more colour reference to change.

functionalDB/chapter.md
```diff
@@ -520,20 +522,21 @@ We use two helper functions to perform the update. *update-attr-modification-tim
...
 
-We can fulfill these requirements through an interface that consumes a database and a set of operations to be performed, and which produces a database whose state reflects the given changes. All of the changes submitted in the batch should be applied through the addition of a _single_ layer. However, we have one small problem: All of the functions we wrote in our low-level API add a new layer to the database. If we were to perform a batch with _n_ operations, we would thus see _n_ new layers added, when what we would really like is to have exactly 1 new layer.   
+The key here is that the layer we want is the _top_ layer that would be produced by performing those updates in sequence. Therefore, the solution is to execute the user’s operations one after another, each creating a new layer. When the last layer is created, we take only that top layer and place it on the initial database (leaving all the intermediate layers to pine for the fjords). Only after we've done all this will we update the database's timestamp.
```
amyrbrown: I'm so sorry, but I think we have to take out the "pine for the fjords" reference here. It will confuse anyone who doesn't get it, and for those who do it doesn't really make sense -- nothing here is dead or apparently dead, only left behind. If you can come up with a clever nerdy "left behind" reference I'll consider it. :)

functionalDB/chapter.md
```diff
@@ -871,7 +876,7 @@ Locating the index of the joining variable is done by *index-of-joining-variable
          collapsed (reduce collapsing-fn metas-seq)] 
      (first (keep-indexed #(when (variable? %2 false) %1)  collapsed)))) 
 ````
-We begin by extracting the metadata of each clause in the query. This metadata is a 3-element vector, each of which is a variable name or *nil*, and reduces them to produce a single variable name or *nil*. If a variable name is produced, this means it appears in all of the metadata vectors at the same index -- i.e. this is the joining variable. We can thus choose the appropriate index by using the mapping described above. 
+We begin by extracting the metadata of each clause in the query. This metadata is a 3-element vector, each element of which is a variable name or `nil`, and reduces them to produce a single variable name or `nil`. If a variable name is produced, this means it appears in all of the metadata vectors at the same index; i.e., this is the joining variable. We can thus choose the appropriate index by using the mapping described above.
```
amyrbrown: The sentence beginning with "This metadata is..." lost me around "reduces them" -- it's not clear *what* reduces *what*. Can you rewrite, and possible break the sentence into two sentences?

functionalDB/chapter.md
```diff
@@ -1,103 +1,104 @@
-# Designing a database like an archaeologist
+<!-- American spelling --> 
+# CircleDB: An Archaeology-Inspired Database
```
amyrbrown: I suspect you wanted to save your excellent name as a surprise at the end, but we typically use the app name in the chapter title. I think the end still has impact because you explain why the name is so perfect.

functionalDB/chapter.md
```diff
@@ -1,103 +1,104 @@
...
+For example, most modern databases implement updates by overwriting old data in-place instead of appending the new data and keeping the old. This mechanism, nicknamed "place-oriented programming" by Rich Hickey [REF], saves storage space but makes it impossible to retrieve the entire history of a particular record. This design decision reflects the computer scientist’s perspective that "history" is less important than the price of its storage. 
 
-If you were to instead ask an archaeologist what should be done with the old data, the answer would be "hopefully, it’s just buried underneath".
+If you were to instead ask an archaeologist what should be done with the old data, the answer would be "hopefully, just bury it underneath".
 
-(Disclaimer: my understanding of a typical archaeologist is based on a few museum visits, several wikipedia articles, and watching the entire Indiana Jones series).
+(Disclaimer: My understanding of the views of a typical archaeologist is based on visiting a few museums, reading several Wikipedia articles, and watching the entire Indiana Jones series.)
```
yoavrubin: This changes from few visits to one museum to undefined number of visits in few museums - I think that we loose some of the contrast here, as I'd like to emphasis that I don't know much about what I'm talking...

functionalDB/chapter.md
```diff
@@ -1,103 +1,104 @@
...
 
-For example, most modern databases implement updates by overwriting old data in-place instead of appending the new data and keeping the old. This mechanism, nicknamed "place oriented programming" by Rich Hickey [REF], saves storage space but makes it impossible to retrieve the entire history of a particular record. This design decision reflects the computer scientist’s perspective that ‘history’ is less important than the price of its storage. 
+For example, most modern databases implement updates by overwriting old data in-place instead of appending the new data and keeping the old. This mechanism, nicknamed "place-oriented programming" by Rich Hickey [REF], saves storage space but makes it impossible to retrieve the entire history of a particular record. This design decision reflects the computer scientist’s perspective that "history" is less important than the price of its storage. 
 
-If you were to instead ask an archaeologist what should be done with the old data, the answer would be "hopefully, it’s just buried underneath".
+If you were to instead ask an archaeologist what should be done with the old data, the answer would be "hopefully, just bury it underneath".
```
yoavrubin: Actually, the question to the archaeologist should have been "where can we find the old data", and not what to do with it

functionalDB/chapter.md
```diff
@@ -1,103 +1,104 @@
...
-`````
+````
 
 A database consists of:
 
-1. Layers of entities, each with its own unique timestamp (rings in Figure 1)
-2. A top-id value which is the next available unique id
-3. The time at which this database was last updated
+1. Layers of entities, each with its own unique timestamp (the rings in Figure 1).
+2. A top-ID value which is the next available unique ID.
```
yoavrubin: top-id is the name of the symbol in the code

functionalDB/chapter.md
```diff
@@ -108,27 +109,28 @@ Similar to entities, creating an attribute is done using the *make-attr* functio
...
 ````
+
 There are a couple of interesting patterns used in this constructor function: 
 
-* We use Clojure’s _Design by Contract_ [ADD REF HERE] pattern to validate that the cardinality parameter is a permissible value
-* We use Clojure’s destructuring mechanism to provide a default value of *:db/single* if one is not given
-* We use Clojure’s metadata capabilities to separate between an attribute's data (name, value and timestamps) and its metadata (type and cardinality). In Clojure, metadata handling is done using the *with-meta* (setting metadata) and *meta* (reading metadata) functions.
+* We use Clojure’s _Design by Contract_ [ADD REF HERE] pattern to validate that the cardinality parameter is a permissible value.
+* We use Clojure’s destructuring mechanism to provide a default value of `:db/single` if one is not given.
+* We use Clojure’s metadata capabilities to distinguish between an attribute's data (name, value and timestamps) and its metadata (type and cardinality). In Clojure, metadata handling is done using the functions `with-meta` (to set) and `meta` (to rea).
```
yoavrubin: to read

functionalDB/chapter.md
```diff
@@ -108,27 +109,28 @@ Similar to entities, creating an attribute is done using the *make-attr* functio
...
-````
+```
 
 ### Storage
 
-So far, we have talked a lot about _what_ we are going to store, without thinking about _where_ we are going to store it. In this chapter, we resort to the simplest storage mechanism, which is storing the data in memory. This is certainly not reliable, but it simplifies development and debugging and allows us to focus on more interesting parts of the program. 
+So far we have talked a lot about _what_ we are going to store, without thinking about _where_ we are going to store it. In this chapter, we resort to the simplest storage mechanism: storing the data in memory. This is certainly not reliable, but it simplifies development and debugging and allows us to focus on more interesting parts of the program. 
 
-We will access the storage via a simple _protocol_ that will make it possible in the future to define additional storage providers for a database owner to select from.
+We will access the storage via a simple protocol that will make it possible to define additional storage providers for a database owner to select from.
```
yoavrubin: I refer here to Clojure's protocol (one of the components in the language), not the general concept of protocol

functionalDB/chapter.md
yoavrubin: I'm fine with starting the title with the name. I would still prefer to keep it "Designing a database like an archaeologist", as it emphasises the main points here, which are the design and the usage of different perspective, as oppose to taking ideas from the domain of archaeology

functionalDB/chapter.md
yoavrubin: would the wikipedia article work for that? http://en.wikipedia.org/wiki/Design_by_contract

functionalDB/chapter.md
yoavrubin: http://www.infoq.com/presentations/Value-Values

functionalDB/chapter.md
yoavrubin: Right, here is a (hopefully) better version:
We begin by extracting the metadata of each clause in the query. This extracted metadata is a 3-element vector, each element in that vector is either a variable name or nil (note that there is no more then one variable name in that vector). Once the vector is extracted, we produce from it (by reducing it) a single value, which is either a variable name or nil.

amyrbrown: Lovely -- thanks! I'll make the change and add it to the PR.

functionalDB/chapter.md
amyrbrown: I changed this to provide satisfying parallels in the structure of the phrases -- they make the text easier and more satisfying to read. I think the message that you don't know what you're talking about is still clear.

functionalDB/chapter.md
amyrbrown: Sorry -- overzealous search-and-replacing. I will fix.

functionalDB/chapter.md
amyrbrown: Fixed, thanks.

functionalDB/chapter.md
amyrbrown: Ah, I see. I rewrote slightly to clarify.

functionalDB/chapter.md
amyrbrown: I'll leave the final decision up to @MichaelDiBernardo --- it will depend on the amount of available space in the table of contents, etc., as well as whether other book chapters will have subtitles and if so, how they are formatted.

functionalDB/chapter.md
amyrbrown: We have a policy against referencing Wikipedia articles (since in effect you're saying, "you can look this up online", which is trivially true). We can leave this unreferenced -- readers can look it up if they need to.

yoavrubin: Sounds good, this is a term that should be familiar for most SW developers anyway

functionalDB/chapter.md
yoavrubin: If it is a real estate issue, I think it is better to remove the CircleDB out, as this conveys no information to the reader about what's in the chapter or why to read it.

functionalDB/chapter.md
amyrbrown: (We decided (in an email discussion) to keep this reference.)

amyrbrown:
I finally made the latest changes to this PR. @MichaelDiBernardo I think it's ready to merge if @yoavrubin agrees.

yoavrubin:
Added a comment on one minor issue. Besides that we still need to decide about the chapter's title and resolve the colours issue.

MichaelDiBernardo:
@yoavrubin I'm finalizing some things on the typesetting front that will inform the color issue, as well as the title length issue.  If it turns out that the title is too long, I agree in this case that we can remove the `CircleDB` name because it's not a project name that would have been known outside of this work (unlike a lot of the projects discussed in earlier AOSA books.) I've been out sick most of this week, but hope to resume stuff on Monday.

functionalDB/chapter.md
```diff
@@ -1,103 +1,104 @@
...
-* Each slice is a labeled artifact (‘a’ through ‘e’)
-* Each artifact has a ‘color’ attribute (where white means that no update was made)
-* Black arrows denote a change in color between layers (e.g., from c.color @t2 to c.color @t0)
-* Light blue arrows are arbitrary relationships of interest between entities (e.g., from ‘b’ to ‘d’)
+* The entire circle is the excavation site.
+* Each ring is a _layer_ (here numbered from 0 to 4).
+* Each slice is a labeled artifact (‘a’ through ‘e’).
+* Each artifact has a ‘color’ attribute (where white means that no update was made).
+* Black arrows denote a change in color between layers (e.g., from c.color @t2 to c.color @t0).
+* Light blue arrows are arbitrary relationships of interest between entities (e.g., from ‘b’ to ‘d’).
```
yoavrubin: "between entities"-> "between artifacts"

functionalDB/chapter.md
```diff
@@ -146,40 +148,37 @@ And here's our in-memory implementation of the protocol, which uses a map as the
    (drop-entity [storage entity] (dissoc storage (:id entity))))
 ````
 
-### Querying our data
+### Querying our Data
```
yoavrubin: How about making the title of this section "Indexing the data" instead of "Querying our data"?

functionalDB/chapter.md
```diff
@@ -726,42 +731,40 @@ A clause in a query is composed of three predicates. The following table defines
...
 
-The *:find* part of the query is transformed into a set of the given variable names:
+The `:find` part of the query is transformed into a set of the given variable names:
 
 ````clojure
 (defmacro symbol-col-to-set [coll] (set (map str coll)))
 ````
 
-The *:where* part of the query retains its nested vector structure. However, each of the terms in each of the clauses is replaced with a predicate according to Table 3. 
+The `:where` part of the query retains its nested vector structure. However, each of the terms in each of the clauses is replaced with a predicate according to Table 3.
```
yoavrubin: The caption of the table was removed, so we need a different way to refer to it

functionalDB/chapter.md
```diff
@@ -726,42 +731,40 @@ A clause in a query is composed of three predicates. The following table defines
...
 3. Executing the plan: Execute the plan and send its results to the next phase.
 4. Unification and reporting: Extract only the results that need to be reported and format them as specified.
 
-#### Phase 1 - Transformation
+#### Phase 1: Transformation
 
 In this phase, we transform the given query from a representation that is easy for the user to understand into a representation that can be consumed efficiently by the query planner. 
 
-The *:find* part of the query is transformed into a set of the given variable names:
+The `:find` part of the query is transformed into a set of the given variable names:
```
yoavrubin: How about adding a sentence before this line that says something like "A query has two parts, the :find part and the :where part, each one transforms in a different way"

functionalDB/chapter.md
MichaelDiBernardo: Yes, I think we could leave the CircleDB out if we want to keep the original title.

functionalDB/chapter.md
MichaelDiBernardo: When I typeset this chapter, I'll use our builtin referencing stuff to make a hard reference to the table.

functionalDB/chapter.md
MichaelDiBernardo: I'll add this during typesetting.

MichaelDiBernardo:
OK, I finally had time to revisit the colour issue.

@yoavrubin, this was my mistake. I didn't realize our print version would be constrained to monochrome way back when we started working on this chapter.

I don't think the second and third diagrams need to change (image1 and image2.) The colors will render as different shades of grey in the print/mobile versions, but the color isn't as important in those diagrams as the physical layering is, and I think the drawn borders make that quite clear. We could then remove all references to color there.

The EAVT/AVET/VEAT/VAET index table (right before the "Database" section) also doesn't seem to benefit much from the color analogy, so I think it can simply be removed without any changes.

The first diagram (image_0.png) is more difficult. I think what we should do here is:

- Reference the colored areas as "shaded' areas, which represent that an attribute was changed (i.e. don't name specific colors)
- Instead of using black and light-blue arrows, use solid arrows for attribute changes and dashed arrows for relationships (or vice-versa)

This means that the only change needed in this diagram is to the arrows. In the web version, the image will be full-color and look fine; in the print versions, it will look good enough to get the point across and no information will be lost.

@yoavrubin, does this work for you? If so, could you make the arrow changes to the first image? I can do all the rest (e.g. the work of changing the text.)

yoavrubin:
How about updating image_0 to use other indicator as a property instead of color, (e.g., greek letters for inner circles, some medieval font for middle circles, and a modern font for the outer layers), maybe it would convey even in a better way the notion of change throughout time (and better fit the archaeological metaphor). For the arrows just use thick / thin lines...

As for image 1 and 2, I agree that the color don't provide any additional info if the table doesn't show colors as well.
@MichaelDiBernardo would that work?

MichaelDiBernardo:
@yoavrubin yes that suggestion is also good, I was just trying to save you as much work as possible :)

functionalDB/chapter.md
```diff
@@ -570,69 +577,67 @@ That transformation occurs in the following transaction call chain:
...
 
-### Graph traversal
+### Graph Traversal
 
-A reference connection between entities is created when an entity’s attribute’s type is *:db/ref*, which means that the value of that attribute is an id of another entity. When a referring entity is added to the database, the reference is indexed at the VAET index.  
-The information found in the VAET index can be leveraged to extract all the incoming links to an entity and is done in the function *incoming-refs*, which collects for the given entity all the leaves that are reachable from it at that index:
+A reference connection between entities is created when an entity’s attribute’s type is `:db/ref`, which means that the value of that attribute is an ID of another entity. When a referring entity is added to the database, the reference is indexed at the VAET index.  
+The information found in the VAET index can be leveraged to extract all the incoming links to an entity. This is done in the `incoming-refs` function, which collects all the leaves that are reachable from the entity at that index:
```
yoavrubin: "which collects all the leaves that are reachable from the entity at that index" => which returns all the entity Ids that refer to the given entity Id by collecting all the VAET index's leaves that are reachable from that Id.

functionalDB/chapter.md
```diff
@@ -641,7 +646,7 @@ The information found in the VAET index can be leveraged to extract all the inco
          filtered-map (if ref-names (select-keys ref-names all-attr-map) all-attr-map)]
       (reduce into #{} (vals filtered-map))))
 ````
-We can also, for a given entity, go through all of it’s attributes and collect all the values of attribute of type :db/ref, and by that extract all the outgoing references from that entity. This is done at the *outgoing-refs* function 
+We can also go through all of a given entity’s attributes and collect all the values of attributes of type `:db/ref`, and by that extract all the outgoing references from that entity. This is done by the `outgoing-refs` function.
 
 ````clojure
 (defn outgoing-refs [db ts ent-id & needed-keys]
```
yoavrubin: in the signature of outgoing-refs, needed-keys was renamed to ref-names

MichaelDiBernardo:
I'm going to merge this now; I'll open another PR with the new image that @yoavrubin sent by email included, as well as the minor changes we've been discussing here.

functionalDB/chapter.md
MichaelDiBernardo: I think that between this general description and the code itself, the meaning is made clear enough.

functionalDB/chapter.md
MichaelDiBernardo: Good catch, thanks.

functionalDB/chapter.md
```diff
@@ -1123,16 +1130,16 @@ We've finally built all of the components we need to to build our user-facing qu
...
+* Supports ACI transactions (durability was lost when we decided to have the data stored in-memory).
+* Supports "what if" interactions.
+* Answers time-related questions.
+* Handles simple datalog queries that are optimized with indexes.
+* Provides APIs for graph queries.
+* Introduces and implements the notion of evolutionary queries.
 
 There are still many things that we could improve: We could add caching to several components to improve performance; support richer queries; and add real storage support to provide data durability, to name a few.
 
 However, our final product can do a great many things, and was implemented in 488 lines of Clojure source code, 73 of which are blank lines and 55 of which are docstrings.
```
yoavrubin: in 488 lines of Clojure source code => 488 lines of code

functionalDB/chapter.md
```diff
@@ -1123,16 +1130,16 @@ We've finally built all of the components we need to to build our user-facing qu
...
+* Handles simple datalog queries that are optimized with indexes.
+* Provides APIs for graph queries.
+* Introduces and implements the notion of evolutionary queries.
 
 There are still many things that we could improve: We could add caching to several components to improve performance; support richer queries; and add real storage support to provide data durability, to name a few.
 
 However, our final product can do a great many things, and was implemented in 488 lines of Clojure source code, 73 of which are blank lines and 55 of which are docstrings. 
 
-Finally, there's one thing that is still missing - a Name. 
-The only sensible option for an in-memory, index optimized, query supporting, library developer friendly, time aware functional database implemented in 360 lines of Clojure code is CircleDB.
```
yoavrubin: implemented in 360 lines of Clojure code => using exactly 360 lines of Clojure code

PR #129, opened 2015-06-21
Crawler's requests need "Host: xkcd.com" header.

ajdavis:
Crawler's requests need "Host: xkcd.com" header.

Looks like XKCD recently installed or reconfigured Varnish, even
HTTP 1.0 requests require the host is specified to make the examples
work.


MichaelDiBernardo:
:)

PR #138, opened 2015-09-05
correct a typo in `cluster/code/run.py`

xyguo:
line 34 of the original file is
```
Request(node, input, req_done).start()
```
it should be `Requester`


MichaelDiBernardo:
Thank you!

PR #139, opened 2015-09-05
Crawler updates

ajdavis:
@asvetlov would you mind reviewing please?

### DIFF
--- a/crawler/code/crawling.py
+++ b/crawler/code/crawling.py
@@ -61,7 +61,7 @@ class Crawler:
         self.q = Queue(loop=self.loop)
         self.seen_urls = set()
         self.done = []
-        self.connector = aiohttp.TCPConnector(loop=self.loop)
+        self.session = aiohttp.ClientSession(loop=self.loop)
         self.root_domains = set()
         for root in roots:
             parts = urllib.parse.urlparse(root)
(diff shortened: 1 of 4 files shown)


crawler/code/crawling.py
```diff
@@ -141,7 +141,7 @@ def parse_links(self, response):
                 text = yield from response.text()
 
                 # Replace href with (?:href|src) to follow image links.
-                urls = set(re.findall(r'''(?i)href=["']?([^\s"'<>]+)''',
+                urls = set(re.findall(r'''(?i)href=["']([^\s"'<>]+)''',
```
asvetlov: Is it intentional?

ajdavis: Yes! This is the commit: https://github.com/ajdavis/500lines/commit/470cba87faa56f9eb941f14f9b1a1507888f7f6c.

The new comic at http://xkcd.com/1572/ has some Javascript like:
```
document.getElementById("id").href=["http://example.com"];
```
... so that reveals a bug in this regex, which thinks the URL is "[". So now I removed the "?" to require an opening quote in the URL.

(Really I should just use BeautifulSoup....)

asvetlov:
LGTM. Let me take a look on unclosed resources after merging the PR

ajdavis:
Thanks @asvetlov. @MichaelDiBernardo this is ready to go.

asvetlov:
@MichaelDiBernardo ping

MichaelDiBernardo:
Thanks!

PR #143, opened 2015-09-12
crawler.markdown fixes for aiohttp.ClientSession.

ajdavis:
Now that the example code has switched from using bare aiohttp.Request
to the connection-pooling aiohttp.ClientSession, I finish updating the
code blocks in the chapter text, and show how to release the connection
back to the pool.


MichaelDiBernardo:
Cool!

PR #145, opened 2015-09-14
Static analysis typeset

MichaelDiBernardo:
While typesetting this chapter, I found a couple of issues that I could use your help with @astrieanna:

- Lines of Julia code that are > 85 cols are not going to wrap in most of our formats (e.g. https://github.com/aosabook/500lines/pull/145/files#diff-45b6ec8a46c682933c0b3fd122d90851R390), so we need to find a way to manually linebreak these. I'm usually fine to do this on my own in most languages, but I'm not aware of what good style for this looks like in Julia. If you show me how you would break up the expression I noted here, I can probably figure it out for the rest of them.

- The disassemblies from `code_typed` (e.g. https://github.com/aosabook/500lines/pull/145/files#diff-45b6ec8a46c682933c0b3fd122d90851R353) are going to be similarly impossible to render without some way to linebreak them at 85 cols. I tried to find a Julia library function that would pretty-print these, but I didn't find anything that didn't completely change the nature of the information. Do you have any suggestions on how to render these within the 85 col limit?


astrieanna:
I can't figure out how to make in-line comments today. :/

There is no real style-guide for Julia, and formatting is inconsistent between packages (and within the standard library, too).  This is what I'd come up with for that expression:
```julia
jumpback = findnext(x-> (typeof(x) == GotoNode && x.label == l)
                        || (Base.is_expr(x,:gotoifnot) && x.args[end] == l),
                    b, i)
```
I tried to make it easy to see the lambda expression as one clump. If this looks hard to read to you, I'd be happy to make the lambda into a separate named function (which should improve both readabiliy and formatting).


I started attempting to format the `code_typed` output better, but I don't think it worked very well. I am not aware of a way to pretty-print the output of code_typed.

@JeffBezanson, do you have an insight into how to render the output of `code_typed` in 85 cols? (e.g. https://github.com/aosabook/500lines/pull/145/files#diff-45b6ec8a46c682933c0b3fd122d90851R353)

MichaelDiBernardo:
That's cool @astrieanna -- thanks for the tip, I think I can take it from here re: normal code formatting.

WRT the `code_typed` stuff -- I wonder if just manually inserting line breaks to mimic the word wrap one would see in an editor at 85 cols would be a practical alternative? I'll try it and maybe send you the PDF output so you can see how it feels.

MichaelDiBernardo:
@astrieanna just sent you a sample PDF typeset of what this would look like (roughly) in the print version. We can follow-up by email.

MichaelDiBernardo:
@astrieanna signed off on this by email.

PR #146, opened 2015-09-14
Copyedits to Byterun Interpreter chapter

amyrbrown:
Here are my copyedits to the Byterun chapter. Nothing major, mostly just typo fixes and formatting changes. 

There are three queries in the PR, on (new) lines 144, 314, and 393.

### DIFF
--- a/interpreter/chapter.txt
+++ b/interpreter/chapter.txt
@@ -139,9 +139,8 @@ Remember that given valid instruction sets, without any changes to our interpret
 
 At this point, we can begin to see how this structure is extensible: we can add methods on the interpreter object that describe many more operations (as long as we have a compiler to hand us well-formed instruction sets).
 
-XXX STOPPED HERE
-
 #### Variables
+
 Next let's add variables to our interpreter. Variables require four pieces: an instruction for storing the value of a variable, `STORE_NAME`; an instruction for retrieving it, `LOAD_NAME`; and a mapping from variable names to values. For now, we'll ignore namespaces and scoping, so we can store the variable mapping on the interpreter object itself. Finally, we'll have to make sure that `what_to_execute` has a list of the variable names, in addition to its list of constants.
 
 ~~~~
@@ -161,11 +160,11 @@ Next let's add variables to our interpreter. Variables require four pieces: an i
                          ("PRINT_ANSWER", None)],
         "numbers": [1, 2],
         "names":   ["a", "b"] }
-~~~
+~~~~
 
-Our new implementation is below. To keep track of what names are bound to what values, we'll add an "environment" dictionary to the __init__ method. We'll also add STORE_NAME and LOAD_NAME. These methods first look up the variable name in question and then use the dictionary to store or retrieve the its value.
+Our new implementation is below. To keep track of what names are bound to what values, we'll add an "environment" dictionary to the `__init__` method. We'll also add STORE_NAME and LOAD_NAME. These methods first look up the variable name in question and then use the dictionary to store or retrieve its value.
 
-The arguments to an instruction can now mean two different things. It can either be an index into the "numbers" list, or it can be an index into the "names" list. The interpreter knows which it should be by checking what the instruction it's executing. We'll break out this logic --- and the mapping of instructions to what their arguments mean --- into a separate method.
+The arguments to an instruction can now mean two different things. They can either be an index into the "numbers" list, or it can be an index into the "names" list. The interpreter knows which it should be by checking what instruction it's executing. We'll break out this logic --- and the mapping of instructions to what their arguments mean --- into a separate method.
 
 ~~~~
 class Interpreter:
@@ -209,7 +208,7 @@ class Interpreter:
                 self.STORE_NAME(argument)
             elif instruction == "LOAD_NAME":
                 self.LOAD_NAME(argument)
-```
+~~~~
 
 Even with just five instructions, the `run_code` method is starting to get tedious. If we kept this structure, we'd need one branch of the `if` statement for each instruction. Here, we can make use of Python's dynamic method lookup. We'll always define a method called `FOO` to execute the instruction called `FOO`, so we can use Python's `getattr` function to look up the method on the fly instead of using the big `if` statement. The `run_code` method then looks like this:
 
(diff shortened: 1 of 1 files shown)


interpreter/chapter.txt
```diff
@@ -131,9 +137,10 @@ Remember that given valid instruction sets, without any changes to our interpret
         "numbers": [7, 5, 8] }
 ~~~~
 
-At this point, we can begin to see how this structure is extensible - we can add methods on the interpreter object that describe many more operations (as long as we have a compiler to hand us well-formed instruction sets).
+At this point, we can begin to see how this structure is extensible: we can add methods on the interpreter object that describe many more operations (as long as we have a compiler to hand us well-formed instruction sets).
+
+#### Variables
 
-### Variables
 Next let's add variables to our interpreter. Variables require four pieces: an instruction for storing the value of a variable, `STORE_NAME`; an instruction for retrieving it, `LOAD_NAME`; and a mapping from variable names to values. For now, we'll ignore namespaces and scoping, so we can store the variable mapping on the interpreter object itself. Finally, we'll have to make sure that `what_to_execute` has a list of the variable names, in addition to its list of constants.
```
amyrbrown: The text says "Variable require four pieces" but the follow list seems to only contain three things: 

- an instruction for storing,
- an instruction for retrieving, and
- a mapping from variable names to values.

Do we need one more thing on that list, or should we change it to "Variables require three pieces"?

interpreter/chapter.txt
```diff
@@ -296,11 +307,11 @@ So far, the interpreter has executed code simply by stepping through the instruc
...
+Python loops also rely on jumping. In the bytecode below, notice that the line `while x < 5` generates almost identical bytecode to `if x < 10`. In both cases, the comparison is calculated and then `POP_JUMP_IF_FALSE` controls which instruction is executed next. At the end of line 4 --- the end of the loop's body --- the instruction `JUMP_ABSOLUTE` always sends the interpreter back to instruction 9 at the top of the loop. When x < 10 becomes false, then `POP_JUMP_IF_FALSE` jumps the interpreter past the end of the loop, to instruction 34. (The instructions `SETUP_LOOP` and `POP_BLOCK` are responsible for setting up and cleaning up the loop, and we'll touch on them later when we discuss blocks.)
```
amyrbrown: Here you say "we'll touch on [the instructions `SETUP_LOOP` and `POP_BLOCK`] later", but they are not specifically referred to in the block section below. Suggest either removing the phrase "and we'll touch on them later", or expanding block discussion below to explicitly refer to the instructions.

interpreter/chapter.txt
```diff
@@ -372,23 +386,28 @@ c  |       (oldest)      | -> data stack: []
...
+When Ned Batchelder and I were working on Byterun, for a long time we had a significant error in our implementation. Instead of having one data stack on each frame, we had just one data stack on the entire virtual machine. We had dozens of tests made up of little snippets of Python code which we ran through Byterun and through the real Python interpreter to make sure the same thing happened in both interpreters. Nearly all of these tests were passing. The only thing we couldn't get working was generators. Finally, reading the CPython code more carefully, we realized the mistake.[FIXME: footnote: My thanks to Michael Arntzenius for his insight on this bug.] Moving a data stack onto each frame fixed the problem.
```
amyrbrown: Changed Artzenius to Arntzenius as per http://www.rntz.net/ -- I assume that's the same person? (!)

interpreter/chapter.txt
```diff
@@ -1,39 +1,43 @@
...
 
-## A Python interpreter
-Before we begin, let's narrow down what we mean by "a Python interpreter." The word "interpreter" can be used in a variety of different ways when discussing Python. Sometimes "interpreter" refers to the Python REPL, the interactive prompt you get by typing just "python" from the command line. Sometimes people use "the Python interpreter" more or less interchangeably with "Python" to talk about executing Python code from start to finish. In this chapter, "interpreter" has a more narrow meaning: it's the last step in the process of executing a Python program.
+### A Python Interpreter
 
-Before the interpreter takes over, Python performs three other steps: lexing, parsing, and compiling. Together, these steps transform the programmer's source code from lines of text into structured _code objects_ containing instructions that the interpreter knows how to understand. The interpreter's job is then to take these code objects and follow the instructions.
```
amyrbrown: Moved this paragraph down below the discussion of interpreters so the paragraph addressing what "interpreter" means comes first.

interpreter/chapter.txt
```diff
@@ -372,23 +386,28 @@ c  |       (oldest)      | -> data stack: []
 k   ---------------------
 ```
 
-At this point, the interpreter is in the middle of the function call to `bar`. There are three frames on the call stack: one for the module level, one for the function `foo`, and one for `bar`. Once `bar` returns, the frame associated with it is popped off of the call stack and discarded.
```
amyrbrown: "off of" is considered too informal for books (although "out of" is fine -- English!)

akaptur: Huh! "out of" doesn't make sense in the context of a stack, but "popped from the call stack" would be fine.

interpreter/chapter.txt
akaptur: :heart: Thank you for this terrific catch

interpreter/chapter.txt
akaptur: I think the list of variable names referenced in "finally" is the fourth thing, but let me check my version history to make sure. Thanks for the catch!

interpreter/chapter.txt
akaptur: Thanks - I'll take a look and see which of these is the better course of action.

interpreter/chapter.txt
akaptur: hmmm, let me look at this in context.

akaptur:
Thanks for the great comments, @amyrbrown! I'll take a look at the ones that require more thought and get back to you as soon as possible.

interpreter/chapter.txt
amyrbrown: Sorry -- my note was confusing. All we need to change here is to remove the "of", to make it "popped off the call stack".

interpreter/chapter.txt
MichaelDiBernardo: + :heart:

MichaelDiBernardo:
Hi,

I'm going to merge this, open a new branch, and address the open comments. 

I'll check with @akaptur before merging, but I'd also like to early-access publish this chapter soon, and I don't think the open comments are stop-ship :)

interpreter/chapter.txt
MichaelDiBernardo: For the record, I think the fourth part was the list of variable names that `what_to_execute` requires, which is the last sentence in the paragraph.

PR #148, opened 2015-09-15
Import module constant

brettcannon:
Else the code would throw a NameError.

### DIFF
--- a/crawler/crawler.markdown
+++ b/crawler/crawler.markdown
@@ -88,7 +88,7 @@ This method not only wastes electricity, but it cannot efficiently await events 
 Python 3.4's `DefaultSelector` uses the best `select`-like function available on your system. To register for notifications about network I/O, we create a non-blocking socket and register it with the default selector:
 
 ```python
-from selectors import DefaultSelector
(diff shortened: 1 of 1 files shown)


MichaelDiBernardo:
This LGTM. @brettcannon, to further answer your question from Twitter, most of the chapters in the book are explaining source that maps directly to the code found in the `code/` directory for its respective project. 

@ajdavis put in the extra work to iteratively evolve his final contribution from first principles, which I think is why some of these snippets are stripped of a bit of context.

ajdavis:
This LGTM too. (Thanks again @brettcannon !) Since there's already an import statement right there, it doesn't bloat the code to add this constant, and it's inconsistent *not* to import EVENT_WRITE. So let's merge this.

But let's not add imports to code blocks everywhere else throughout the text. It'd distract readers from main discussion. They can refer to the code in the repo, which is complete and executable, if they want. But the code itself is not the point. There are much better web crawlers than this one, and better async frameworks. The code's only purpose is to illustrate principles about async I/O and coroutines.

MichaelDiBernardo:
Yes, I agree completely.

PR #149, opened 2015-09-15
Get the URL from the instance and not some global

brettcannon:


ajdavis:
LGTM, @MichaelDiBernardo we should merge this before publication.

MichaelDiBernardo:
Thanks @brettcannon!

PR #150, opened 2015-09-15
Make Crawler.loop actually exist

brettcannon:
Otherwise the attribute seems to never be set.

### DIFF
--- a/crawler/crawler.markdown
+++ b/crawler/crawler.markdown
@@ -797,7 +797,7 @@ We collect the workers' shared state in a crawler class, and write the main logi
 ```python
 loop = asyncio.get_event_loop()
 
-crawler = crawling.Crawler(loop, 'http://xkcd.com',
+crawler = crawling.Crawler('http://xkcd.com',
                            max_redirect=10)
 
 loop.run_until_complete(crawler.crawl())
@@ -807,7 +807,7 @@ The crawler begins with a root URL and `max_redirect`, the number of redirects i
 
 ```python
 class Crawler:
-    def __init__(self, loop, root_url, max_redirect):
+    def __init__(self, root_url, max_redirect):
         self.max_tasks = 10
         self.max_redirect = max_redirect
         self.q = Queue()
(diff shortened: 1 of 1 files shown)


ajdavis:
Thanks again @brettcannon ! In this case I see the problem you want to solve but disagree with the method a little, since it [diverges from the actual code so much](https://github.com/aosabook/500lines/blob/71e0672/crawler/code/crawling.py#L50-L54). Here's the class in crawling.py:
```
class Crawler:
    def __init__(self, roots,
                 exclude=None, strict=True,
                 max_redirect=10, max_tries=4,
                 max_tasks=10, *, loop=None):
        self.loop = loop or asyncio.get_event_loop()
```
How about in crawler.markdown, we do?:
```
loop = asyncio.get_event_loop()

crawler = crawling.Crawler('http://xkcd.com',
                           max_redirect=10,
                           loop=loop)

loop.run_until_complete(crawler.crawl())
```
So just like you have it, but with `loop` passed as a final keyword arg instead of an initial positional arg. And then, simply:
```
class Crawler:
    def __init__(self, root_url, max_redirect, loop):
        self.loop = loop
        # ... and the rest of that code example ...
 ```
This is a reasonable compromise, it stays closer to the code in crawling.py but still doesn't weigh down the text with irrelevant details.

**Comment by brettcannon** [comment]
@ajdavis Done! PTAL

**Comment by asvetlov** [comment]
BTW I believe original crawler should have signature like
```
class Crawler:
    def __init__(self, roots, *,
                 exclude=None, strict=True,
                 max_redirect=10, max_tries=4,
                 max_tasks=10, loop=None):
                     ...
```

I mean all parameters except `roots` should be keyword-only.

**Comment by ajdavis** [comment]
@asvetlov I see that idiom a lot in new code in the Python 3 standard library. Is it fair to say that it's now deprecated to declare parameters with default values that can be either positional or keyword? That is, this is now Considered Harmful?:
```
def foo(a=1, b=2):
    pass
```
... and in Python 3 such a function should *always* be declared like:
```
def foo(*, a=1, b=2):
   pass
```

brettcannon:
Keyword-only arguments are typically used in cases where providing the argument by name would be ambiguous. For instance, if you see code like `some_func(1, 2, 3, True)` do you know what `True` means? But if the call was `some_func(1, 2, 3, sum=True)` you have a much better idea. So it doesn't directly have anything to do with default arguments as much as whether context is necessary.

Keyword-only arguments are also fantastic for expanding an API that previously existed where you don't want people using the old version of the API to accidentally succeed in calling something when they pass an extra argument in by accident.

ajdavis:
Interesting.

As for this PR, let's merge it, I think this reads well and keeps the user focused on what matters. Thank you. =)

MichaelDiBernardo:
Thanks all.

PR #152, opened 2015-09-21
FunctionalDB copy-edit changes (part 2)

MichaelDiBernardo:
Responded to image changes + other unaddressed comments from https://github.com/aosabook/500lines/pull/124.

@yoavrubin, can you take one last look? I think this is good to go to typesetting now.


functionalDB/functionalDB.markdown
```diff
@@ -585,50 +585,50 @@ At this point we can see that with few minor tweaks, we can also provide a way t
...
 (transact db-conn  (add-entity e1) (update-entity e2 atr2 val2 :db/add))  
-````
+```
 changes into: 
-````clojure
+```clojure
 (_transact db-conn swap! (add-entity e1) (update-entity e2 atr2 val2 :db/add))
-````
+```
 which becomes:
```
yoavrubin: which eventually becomes

functionalDB/functionalDB.markdown
```diff
@@ -585,50 +585,50 @@ At this point we can see that with few minor tweaks, we can also provide a way t
...
-````
+```
 which becomes: 
-````clojure
+```clojure
 (swap! db-conn transact-on-db [[add-entity e1][update-entity e2 atr2 val2 :db/add]])
-````
+```
 
 The what-if user call:
```
yoavrubin: To keep it structured the same way as transaction, how about changing this sentence to be "For What-if, the user call"

functionalDB/functionalDB.markdown
```diff
@@ -668,14 +668,14 @@ Our data model is based on accumulation of facts (i.e., datoms) over time. For t
 ### Query Language
```
yoavrubin: a subset of the Datalog language from the => a subset of the Datalog-based query language from the

functionalDB/functionalDB.markdown
MichaelDiBernardo: I think 'which becomes' is terser and communicates essentially the same point.

functionalDB/functionalDB.markdown
MichaelDiBernardo: We don't need to repeat "query language" here because it was already identified as such when we said " A commonly used query language influenced by logic programming is _Datalog_ which..."

MichaelDiBernardo:
@yoavrubin I've changed the title back. I'll check how long it is in the table of contents when typesetting is done; if we need to shorten it, we'll use the title @amyrbrown suggested. I've also addressed your other comments. 

I'm going to merge this now so that I can start typesetting it; we can do a last pass when that PR is opened.

yoavrubin:
Thanks @MichaelDiBernardo , looks good.

PR #154, opened 2015-09-28
FunctionalDB typeset [WIP]

MichaelDiBernardo:


functionalDB/functionalDB.markdown
```diff
@@ -1,4 +1,4 @@
-title: Designing a Database Like an Archaeologist
+title: An Archaeology-Inspired Database
```
yoavrubin: I guess that the original name was too long?

MichaelDiBernardo: screen shot 2015-09-27 at 9 43 28 am

:)

MichaelDiBernardo: It was too long to render in chapter headers and Table of Contents, yes. I did change the second headline to be identical to your original title, so that your point is still made early in the chapter.

yoavrubin: I think you've nailed it with that addition of the second headline, much smoother structure now

MichaelDiBernardo:
@yoavrubin this looks good for the HTML early-access release. However, for the PDF/print format, we have some line-length problems :)

We're going to need to restrict all code lines to be < 85 columns, including leading whitespace. In the HTML version this doesn't matter because the code blocks have horizontal scrolling. 

I don't know enough about spacing/alignment conventions in Clojure to feel comfortable doing this. Do you mind giving it a shot? I can identify all the places where this happens if it makes your life easier, but the way I found them was just by getting my editor to highlight line-endings that exceeded the 85 column limit, and then I just looked at all the code blocks.

There's one more width problem that I'll highlight with a specific column.

Like I said, this isn't super urgent because we can ship this as-is for the early-access release.

functionalDB/functionalDB.markdown
```diff
@@ -843,6 +907,28 @@ and the following structure for the `:where` part. (Each cell in the _Predicate
...
+{\footnotesize
+\rowcolors{2}{TableOdd}{TableEven}
+\begin{tabular}{lll}
+\hline
+\textbf{Query Clause} & \textbf{Predicate Clause} & \textbf{Meta Clause} \\
+\hline
+\verb|[?e  :likes "pizza"]| & \verb|[#(= % %)  #(= % :likes)  #(= % "pizza")]| & \verb|["?e" nil nil]| \\
+\verb|[?e  :name  ?nm]| & \verb|[#(= % %)  #(= % :name) #(= % %)]| & \verb|["?e" nil "?nm"]| \\
+\verb|[?e  :speak "English"]| & \verb|[#(= % %) #(= % :speak) #(= % "English")]| & \verb|["?e" nil nil]| \\
+\verb|[?e  :birthday (birthday-this-month? ?bd)]| & \verb|[#(= % %) #(= % :birthday) #(birthday-this-month? %)]| & \verb|["?e" nil "?bd"]| \\
```
MichaelDiBernardo: This row is way too wide in the print version. Would it hurt the clarity much here to change all instances of `birthday` to `bday`, and to name the predicate `bday-this-mo`? (We'd probably have to make this change in the other tables / examples also for consistency.) If you have a better suggestion, I'm open to it :)

MichaelDiBernardo: screen shot 2015-10-01 at 8 27 00 am

yoavrubin: Don't think that there would be any harm with that change

yoavrubin:
@MichaelDiBernardo  I'll make the needed alignment changes in the chapter - I'll create a PR handling it once you merge all your changes, would that be ok ?

MichaelDiBernardo:
@yoavrubin I'll merge now and make the table changes I mentioned in a downstream commit. Feel free to branch from master :)

PR #155, opened 2015-10-02
adapting length of code lines to 85 columns

yoavrubin:
@MichaelDiBernardo please let me know if there are still lines that need trimming


MichaelDiBernardo:
Thanks! This looks great, I don't see any other lines that require wrapping.

PR #156, opened 2015-10-07
Fix small objmodel problems

cfbolz:
Hi @MichaelDiBernardo,

I spotted two problems in the html version of the objmodel chapter:
 - the svg export went wrong somehow and the arrowheads got lost.
 - one footnote had a typo in the markdown syntax that makes it not be a footnote in html


MichaelDiBernardo:
Thank you!

PR #157, opened 2015-10-11
crawler: typo

perrinjerome:
It seems to me that having "it" and "gen" twice at the end of this sentence is a typo

The `yield from` statement is a frictionless channel, through which values flow in and out of `gen` until *it* `gen` completes.


MichaelDiBernardo:
Thank you!

ajdavis:
Thanks @perrinjerome

PR #161, opened 2015-10-21
A few additions to web server chapter

MichaelDiBernardo:


web-server/chapter.md
```diff
@@ -516,6 +518,7 @@ we need to modify `handle_error` and `send_content` as follows:
     # Handle unknown objects.
     def handle_error(self, msg):
         content = self.Error_Page.format(path=self.path, msg=msg)
+        # This should be sending more than just 404?
```
MichaelDiBernardo: @gvwilson I think we need to mention here that `404` isn't a sufficient error code for all the cases we're handling -- especially since later, when we implement cgi, we still use this to signal all errors when an underlying problem in the called script should almost certainly be a `500`. 

I'm happy enough just to say "this isn't good enough, but we're going to stick to using 404 throughout this chapter so we can focus on more interesting things." Does that work for you? If so, I'll just add it in.

MichaelDiBernardo:
@gvwilson I've made a few minor additions, and I have an open question for you (see other comment). If I don't hear back by 26-Oct, I'll assume you're OK with it and just go ahead :sunglasses:

web-server/chapter.md
gvwilson: :+1:

gvwilson:
Looks good to me - thanks for the fixes.

PR #169, opened 2015-12-01
Dagoba typeset

MichaelDiBernardo:


MichaelDiBernardo:
@dxnn: The biggest changes in this were:

- Typesetting all the footnotes
- Formatting the code to 85 cols, which is as wide as we can go in the book.

I made some pretty arbitrary formatting decisions with the code -- I was specifically not consistent with when I left the comments on the same line, and when I tugged them up a line. I also wasn't particularly consistent with how I broke up really long code lines (usually chained query methods.) 

I did like the visual style of having them all on the same line, but unfortunately that won't work in print. If you'd like to give a go at being more consistent / doing something altogether different, I'm happy to take suggestions.

PR #175, opened 2015-12-18
Fix spacing between commands

Deborah-Digges:


MichaelDiBernardo:
This is a bit clearer, thanks. (They should both work, though.)

PR #179, opened 2015-12-22
Dagoba typeset

dxnn:
Mostly line alignment and whitespace changes.


dagoba/dagoba.markdown
```diff
@@ -84,31 +84,25 @@ Given a list of vertices we then reduce over the edges, adding an edge's parent
...
-parents  = function(x) { return E.reduce( 
+parents  = function(x) { return E.reduce(
   function(acc, e) { return ~x.indexOf(e[1]) ? acc.concat(e[0]) : acc }, [] )}
-children = function(x) { return E.reduce( 
+children = function(x) { return E.reduce(
   function(acc, e) { return ~x.indexOf(e[0]) ? acc.concat(e[1]) : acc }, [] )}
 ```
 
-Now we can say something like: 
+Now we can say something like: `children(children(parents(parents([8]))))`. It reads backwards and gets us lost in silly parens, but is otherwise pretty close to what we wanted. Take a minute to look at the code. Can you see any ways to improve it?
```
MichaelDiBernardo: Thanks for shortening this. I reserve the right to put it back in its own block, though :) In LaTeX, the code between the `ticks` is forced to be rendered as one token, which makes it hard to fit anything longer than a single keyword on the same line with normal text.

MichaelDiBernardo:
LGTM, thanks!

dagoba/dagoba.markdown
dxnn: Ah, I see. No problem. I've propagated that change through to other places as well now -- I'll make a new PR for that.

PR #185, opened 2016-01-17
[catechism]move raise logic to matcher constructors 

conndots:
I move the raising logic to the constructors of matchers, rather than staying in the subject_wrapper. Because when it should end matching and raise an excaption is a logic belonging to the matcher itself like the change matcher to have following calls to do and raise them then.


MichaelDiBernardo:
Hi, thanks for this, but the catechism chapter isn't going to be in the final book.

PR #190, opened 2016-02-11
Backporting typesetting changes into Dagoba copyedits

MichaelDiBernardo:
For the sake of speed, copyedits by @amyrbrown were done on a version of the chapter that didn't have a bunch of the footnote + code formatting stuff that @dxnn and I worked on downstream. This resulted in a sort-of-nasty merge. I also took this opportunity to actually name the footnotes, instead of using `[^footnoteA]` , `[^footnoteB]`, etc.

@dxnn: Can you take a look at this using a local diff client to see if there's anything you think requires revisiting? Alternately, you can just take a look at the raw, or just trust us on this one :)

The main changes that will be seen in this diff are:

- The actual copyedits made
- Backporting footnote formatting changes made in a later typeset
- Backporting code formatting changes made in a later typeset
- Renaming the footnotes to descriptive tags.

### DIFF
--- a/dagoba/dagoba.markdown
+++ b/dagoba/dagoba.markdown
@@ -1,57 +1,58 @@
-title: Dagoba: an in-memory graph database
-author: Dann Toliver
+# Dagoba: an in-memory graph database
 
-_[Dann](https://twitter.com/dann) enjoys building things, like programming languages, databases, distributed systems, communities of smart friendly humans, and pony castles with his two year old._
+> "When we try to pick out anything by itself we find that it is bound fast by a thousand invisible cords that cannot be broken, to everything in the universe." 
+> --- John Muir 
 
-> "When we try to pick out anything by itself we find that it is bound fast by a thousand invisible cords that cannot be broken, to everything in the universe."
-> -- John Muir
+&nbsp;
 
-> "What went forth to the ends of the world to traverse not itself, God, the sun, Shakespeare, a commercial traveller, having itself traversed in reality itself becomes that self."
-> -- James Joyce
+> "What went forth to the ends of the world to traverse not itself, God, the sun, Shakespeare, a commercial traveller, having itself traversed in reality itself becomes that self." 
+> --- James Joyce 
+
+## Prologue
 
 A long time ago, when the world was still young, all data walked happily in single file. If you wanted your data to jump over a fence, you just set the fence down in its path and each datum jumped it in turn. Punch cards in, punch cards out. Life was easy and programming was a breeze.
 
-Then came the random access revolution, and data grazed freely across the hillside. Herding data became a serious concern -- if you can access any piece of data at any time, how do you know which one to pick next? Techniques were developed for corralling the data by forming links between items [^footnoteA], marshaling groups of units into formation through their linking assemblage. Questioning data meant picking a sheep and pulling along everything connected to it.
+Then came the random access revolution, and data grazed freely across the hillside. Herding data became a serious concern: if you can access any piece of data at any time, how do you know which one to pick next? Techniques were developed for corralling the data by forming links between items [^items], marshaling groups of units into formation through their linking assemblage. Questioning data meant picking a sheep and pulling along everything connected to it. 
 
-Later programmers departed from this tradition, imposing a set of rules on how data would be aggregated[^footnoteB]. Rather than tying disparate data directly together they would cluster by content, decomposing data into bite-sized pieces, collected in kennels and collared with a name tag. Questions were declaratively posited, resulting in accumulating pieces of partially decomposed data (a state the relationalists refer to as "normal") into a frankencollection returned to the programmer.
+Later programmers departed from this tradition, imposing a set of rules on how data would be aggregated[^relationaltheory]. Rather than tying disparate data directly together they would cluster by content, decomposing data into bite-sized pieces, collected in pens and collared with name tags. Questions were posed declaratively, resulting in accumulating pieces of partially decomposed data (a state the relationalists refer to as "normal") into a frankencollection returned to the programmer.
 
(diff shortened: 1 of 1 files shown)


amyrbrown:
I usually leave comments directly on the diff'd file on Github, but Github won't show a diff because there are too many changes, so I'll put them here. Line numbers refer to the file I was working on, so they probably are slightly off by now.

Title: I wonder if we should put a footnote explaining the provenance of this name, so people don't think you misspelled "Dagobah".

Line ~21: Will readers know what "CAP-wielding" refers to?

Footnote beginning "Actually, we only used this underscore technique": This footnote doesn't make sense now that there is no comment on the first line, and it didn't really make sense before because you can just use spaces to line up comments -- right? Sugest clarifying or removing.

Line ~438: "There are two issues keeping us from doing so." There's only a direct reference to one issue. Suggest adding something like "The second issue is" before the second issue is introduced.

Line ~498: "whose weight in skippund is greater than their height in fathoms". Would be better if this were rewritten with more conventional units -- this material is pretty hard already and we want to lighten the cognitive load where we can. (Think of reading this with English as a second language.) Sorry. :( I know you're trying to make it more fun!

Line ~1271: "Scoping the freezing/thawing pipes may be tricky" --- I'm assuming that the reference to freezing/thawing is a continuation of the suspended animation metaphor, but it took me a minute. Suggest rewriting without reference to the method of suspension, e.g. "Scoping the suspending/unsuspending pipes may be tricky"

Thank you! I enjoyed editing this chapter, @dxnn

dxnn:
Hi Amy,

Thank you for the lovely copy editing job! I was shocked, slightly saddened and thoroughly delighted by the number of e.g. missing words and pluralizations you found and fixed. 

There's quite a lot to go through here -- in addition to replying to your comments I've also assembled a list of changes to revert or revisit (a few technical, a few personal opinion). I'm not sure what standard protocol is here, but maybe the easiest path forward would be to accept this pull request, and then I'll make my edits as a separate PR, so we can have our web-diff and inline comments back.

Comments on your comments:

Title footnote: Good idea! The origin story actually involves a few different aspects (Directed Acyclic Graphs, chocolate, planets, temples, and so on). I'll weave something together for this.

Line 21ish: It's quite common in certain circles -- I suspect most folks will be somewhat familiar with it. (It's called the CAP Theorem, for reference.)

"Actually, we only used this underscore technique": I agree that it doesn't make sense any more, but I'm relectant to delete it as I've had a couple of comments from readers who appreciated that bit. I'll rework it so it makes more sense. (The missing variable names were quite long, so they would have caused other comments to run off the end of the page if all the comments were aligned.)

Line ~438: Good idea. I'll rework that section to highlight the second issue.

Line ~498: I'll convert this so the esoteric measurements are in a footnote instead of the main text. This might also be a way to sneak Kirby/Volstagg back in (they fell out in an earlier edit and I've been missing them).

Line ~1271: That makes sense. I'll replace it with "suspending/unsuspending"

Comments on changes, and a couple disclaimers: I'm not sure who made the following edits, since there are so many merges here, so I may well be commenting on things I myself changed in the past. Also, I suspect I can sometimes be a bit too precious with joke material, and I'm willing to listen when that happens. 

Line numbers below refer to line numbers in this pull request. I've included notes to myself for the followup cleanup as well -- let me know if you have any questions.

30/32: This was actually a Python reference (Monty, not Guido). I'd prefer to revert, but could otherwise rework.
55: If we take the columns out of alignment then I'll cut this differently so the lines line up better (or revert to column alignment, either way is fine with me)
114: This comment is now out of alignment with the below comments, but also serves as a reminder to me to ensure proper alignment throughout to document
173: Change to "`this.vertices.length` here"
193: Revert to 2 spaces instead of 4
200: Revert comment wording to "add to this edge's out vertex's out edges"
282: Work 'strict' back in here -- possibly "are strict in their insistence" (it's important for the joke)
550: Can we restore "so we walk all the way back to the primeval cow Auðumbla"? She's referenced by name a few lines later (561), and it helps set the context more as "we're going to go back to the beginning", rather than "we're going to go back a long but arbitrary distance".
899: Restore memoization (it doesn't spell check, but is the correct term)
1063: We lost this sentence at the end of the paragraph: 'Note that aliases run with a high priority value of 100, because there's no way to catch missing methods in JavaScript, so we have to perform the transformation before any aliased methods can be invoked.'
1135: The interesting part of this footnote for me was the first half -- suggest we either restore it or cut the whole footnote. "Though only in operation for 18 months due to the arrival of the transcontinental telegraph and the outbreak of the American Civil War, the ..." 
1207: I'm willing to lose 'insufferably persnickety', but perhaps we can compromise on 'rather persnickety'?
1248: Revert to "Blob", as that's the capitalization in JS
1300: Perhaps a colon here instead? "...every pathway is present in the graph; if we're missing an entry..."
1316: Maybe restore stolid/yolo modifiers?
1325: Change this to something like "Loki having 1,000 new grandkids doesn't increase Thor's degree at all"
1330: Add Amy to the Acknowledgements

I was trying to be careful about labelling strings with 'string' as opposed to `code`, so in the section where I'm referring to 'done' and 'as' and 'pull' as signals I used that convention -- they've been switched to `done` and `as` and `pull`. I'm happy to conform to house style on this, but figured I would mention the reasoning, as I think the disambiguation might be useful.

I'm happy to make all the changes suggested above, and I should have time in the next few days to do so. Let me know how you'd like to proceed. 

Thanks again for the amazing work on this article! I'm really happy with how it's all come together. ^_^

MichaelDiBernardo:
@dxnn thanks -- let's merge this version, and then have you open a PR with your final edits which references this one. 

Merging now!

amyrbrown:
I'm glad you're happy with the edits, @dxnn! Thank you for the thoughtful responses. Once you issue your PR, I'll have a look at your reversions and comments in context.

PR #192, opened 2016-02-19
Edits to the web server chapter.

amyrbrown:
As usual, editing @gvwilson's writing is boring as hell with almost nothing to change or fix. SIGH. I did manage to find a couple of things to complain about -- see comments in the "Files changed" view.

### DIFF
--- a/web-server/web-server.markdown
+++ b/web-server/web-server.markdown
@@ -1,7 +1,7 @@
 title: A Simple Web Server
 author: Greg Wilson
 
-_Greg Wilson is the founder of Software Carpentry, a crash course in computing skills for scientists and engineers.  He has worked for 30 years in in both industry and academia, and is the author or editor of several books on computing, including the 2008 Jolt Award winner "Beautiful Code" and the first two volumes of "The Architecture of Open Source Applications". Greg received a PhD in Computer Science from the University of Edinburgh in 1993._
+_Greg Wilson is the founder of Software Carpentry, a crash course in computing skills for scientists and engineers.  He has worked for 30 years in both industry and academia, and is the author or editor of several books on computing, including the 2008 Jolt Award winner *Beautiful Code* and the first two volumes of *The Architecture of Open Source Applications*. Greg received a PhD in Computer Science from the University of Edinburgh in 1993._
 
 ## Introduction
 
@@ -36,7 +36,7 @@ such as `174.136.14.108`;
 the Domain Name System (DNS) matches these numbers to symbolic names like `aosabook.org`
 that are easier for human beings to remember.
 
-A port number is just a number in the range 0-65535
+A port number is a number in the range 0-65535
 that uniquely identifies the socket on the host machine.
 (If an IP address is like a company's phone number,
 then a port number is like an extension.)
--- a/web-server/web-server.markdown
+++ b/web-server/web-server.markdown
@@ -1,7 +1,7 @@
 title: A Simple Web Server
 author: Greg Wilson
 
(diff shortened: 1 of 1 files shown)


web-server/web-server.markdown
```diff
@@ -1,7 +1,7 @@
 title: A Simple Web Server
 author: Greg Wilson
 
-_Greg Wilson is the founder of Software Carpentry, a crash course in computing skills for scientists and engineers.  He has worked for 30 years in in both industry and academia, and is the author or editor of several books on computing, including the 2008 Jolt Award winner "Beautiful Code" and the first two volumes of "The Architecture of Open Source Applications". Greg received a Ph.D. in Computer Science from the University of Edinburgh in 1993._
+_Greg Wilson is the founder of Software Carpentry, a crash course in computing skills for scientists and engineers.  He has worked for 30 years in both industry and academia, and is the author or editor of several books on computing, including the 2008 Jolt Award winner *Beautiful Code* and the first two volumes of *The Architecture of Open Source Applications*. Greg received a PhD in Computer Science from the University of Edinburgh in 1993._
```
amyrbrown: @MichaelDiBernardo If this bio is printed in italics, the titles of the books should be in roman, and vice versa.

web-server/web-server.markdown
```diff
@@ -110,16 +110,17 @@ and other codes have other meanings.
...
 and the server doesn't remember anything between one request and the next.
 If an application wants to keep track of something like a user's identity,
 it must do so itself.
+
 The usual way to do this is with a cookie,
-which is just a short character string that the server sends to the client,
+which is a short character string that the server sends to the client,
 and the client later returns to the server.
-
 When a user signs in,
```
amyrbrown: "signs in" here implies that the user is actually logging in to something, but servers don't need you to log in in order to create cookies -- suggest changing to "When a user visits a page," or something similar.

web-server/web-server.markdown
```diff
@@ -170,7 +172,7 @@ Python comes with such a library called `urllib2`
 but it exposes a lot of plumbing that most people never want to care about.
 Instead,
 we recommend using the [Requests](https://pypi.python.org/pypi/requests) library.
```
amyrbrown: Who is "we" here? Since there's only one author suggest using "I" to avoid ambiguity, or rewriting to use passive voice.

amyrbrown:
I also have comments on the images.

`http-cycle.png`: I suggest removing the fifth and sixth arrows (from the left), because step 5 (client finds references to images) does not *cause* the server to close the connection. I would move the "Server closes connection" text to the end of step 4. Perhaps an arrow from step 5 to step 6?

Also need a capital on "Client" in steps 1 and 5, and utoronto is misspelled in step 6.

`http-request.png` and `http-response.png`: I'm not clear why these are images rather than text (unless it's to show the CR/LF/spaces more clearly). I think this information would be clearer as text -- the boxes and the font make the snippets of text seem more like objects, not text.

web-server/web-server.markdown
MichaelDiBernardo: The bios aren't going to be included in the chapters like this in print, I just haven't gotten around yet to putting the target tags around them yet.

MichaelDiBernardo:
@amyrbrown I can cover these edits for @gvwilson since there are so few.

MichaelDiBernardo:
I think the request/response images should stay as-is because it's not particularly easy to explicitly show CRLF in a comprehensible way in plaintext. I appreciate the comment about the spaces looking like objects, but I think we come out ahead by keeping them.

web-server/web-server.markdown
amyrbrown: I figured! I just wanted to flag that I changed the title styling from quotations to emphasis.

PR #195, opened 2016-02-28
Event-driven web framework edits

MichaelDiBernardo:
@Inaimathi: I've reworked this chapter so that it's split into two major sections -- the first focuses on building a webserver, and the second focuses on expanding that server into a framework that makes it easier for application programmers to build their own apps. You already had this distinction in your previous draft; my goal here was to use that as a skeleton + outline to guide the development of the chapter.

I moved a lot of things around, and probably removed several things. The biggest change structurally was that I inverted the macro explanations so that we see definitions before expansions.

There are a few things I need help with:

1. I had some questions that I couldn't (or didn't!) answer when doing this -- can you respond to those? (I think I can probably do the defparameter one on my own.)

2. It's likely that some of my changes introduced errors -- by which I mean I may have tried to clarify or move something, and in doing so, written something that is untrue or incorrect. Can you keep an eye out for this?

3. All Lisp code (including macro expansions!) needs to be formatted so that it's no more than 85 columns wide (this includes the leading whitespace.) I'm not 100% sure what tasteful wrapping conventions look like in Lisp, so I was loathe to do this. My results when I pasted into Emacs and force-formatted weren't pretty.

Let's get the first two points figured out on this PR. Once that's done, I can merge, and you can open another PR with the code formatting changes for point 3. 

I think that's it :)


event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+            | Addition | Subtraction | Multiplication |
+    ----------------------------------------------------------
+    Integer |          |             |                |
+	-----------------------------------------------------
+	Real    |          |             |                |
+	-----------------------------------------------
+	Complex |          |             |
+```
+
+NOTE: Is this a misleading analogy? If we consider the specific case of these being binary operators (let's not handle things like (+ x y z w)) isn't it more like an N-d table, or for simplicity:
```
MichaelDiBernardo: Note

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+- loops to read characters from the input stream, and 
+- returns the last character it read if it has read all of the available input. 
+- It also tracks any `\r\n\r\n` sequences so that we can later detect complete requests. 
+- Finally, any error results it returns an `:eof` to signal that `process-ready` should discard this particular connection.
+
+The `buffer` type is a CLOS _class_. Classes in CLOS let us define a type with fields called `slots`. We don't see the behaviours associated with `buffer` on the class definition, because (as we've already learned), we do that using generic functions like `buffer!`.
+
+`defclass` does allow us to specify getters/setters (`reader`s/`accessor`s), and slot initializers; `:initform` specifies a default value, while `:initarg` identifies a hook that the caller of `make-instance` can use to provide a default value. 
+
+NOTE: Generic functions can specialize on types that aren't classes, right? The example of `port` from the first method definition didn't seem like it was a class. I get the impression that classes are just giving us some syntactic sugar over common operations on structs.
```
MichaelDiBernardo: Note

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+
+```lisp
+	(defparameter *channels* (make-hash-table))
+
+	(defmethod subscribe! ((channel symbol) (sock usocket))
+	  (push sock (gethash channel *channels*))
+	  nil)
+```
+
+NOTE: What is defparameter?
```
MichaelDiBernardo: Note

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+		     when (ignore-errors 
+			    (write! msg sock)
+			    (force-output (socket-stream sock))
+			    sock)
+		     collect it))))
+```
+
+In `publish!`, we call `write!` to actually write an `sse` to a socket. We'll also need a specialization of `write!` on `response`s to write full HTTP responses as well. Let's handle the HTTP case first.
+
+NOTE -- Why aren't we worried about `write-ln` blocking?
```
MichaelDiBernardo: Note

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+```lisp
+    (defmacro define-handler ((name &key (close-socket? t) (content-type "text/html")) (&rest args) &body body)
+      (if close-socket?
+          `(bind-handler ,name (make-closing-handler (:content-type ,content-type) ,args ,@body))
+          `(bind-handler ,name (make-stream-handler ,args ,@body))))
+```
+
+It delegates to three other macros (`bind-handler`, `make-closing-handler`, `make-stream-handler`) that we will define later. `make-closing-handler` will create a handler for a full HTTP request/response cycle; `make-stream-handler` will instead handle an SSE message. The predicate `close-socket?` distinguishes between these cases for us. The backtick and comma are macro-specific operators that we can use to "cut holes" in our code that will be filled out by values specified in our Lisp code when we actually use `define-handler`.
+
+NOTE: I think maybe inverting the result of this function and calling it `is-sse?` or `is-stream` might be clearer than the name `close-socket?`
```
MichaelDiBernardo: Note

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+
+Now that we've motivated our architectural decision and decided on a mechanism for simulating bidirectional communication between clients and server, let's get started on building our web framework. We'll start by building a relatively "dumb" server first, and then we'll extend it into a web-application framework that lets us focus on _what_ our heavily-interactive program needs to do, and not _how_ it is doing it.
+
+## Building an Event-Driven Web Server
+
+### The Event Loop
+
+The core of every event-driven program is the _event loop_, which looks something like this: 
+
+```lisp
```
Inaimathi:
```
(defmethod start ((port integer))
  (let ((server (socket-listen 
		 usocket:*wildcard-host* port
		 :reuse-address t
		 :element-type 'octet))
	(conns (make-hash-table)))
    (unwind-protect
	 (loop (loop for ready
		  in (wait-for-input
		      (cons server (alexandria:hash-table-keys conns))
		      :ready-only t)
		  do (process-ready ready conns)))
      (loop for c being the hash-keys of conns
	 do (loop while (socket-close c)))
      (loop while (socket-close server)))))
```

(As a note, 85 columns is pretty aggressively narrow for Lisp code, but I'll do my best)

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+- dispatches on the type of its arguments to figure out which method body should be run, and
+- runs the appropriate function.
+
+You can think about it as a giant table with "Type Name" down the first column and "Operation" across the first row, as in Table FIXME.
+
+```
+            | Addition | Subtraction | Multiplication |
+    ----------------------------------------------------------
+    Integer |          |             |                |
+	-----------------------------------------------------
```
Inaimathi: indentation

event-web-framework/draft.markdown
Inaimathi: Sort of. We have `struct`s in CL too; the main difference between a `class` and a `struct` is that `struct`s don't implement any kind of inheritance. You can see an empty list following the class-name of each definition in this chapter; that can designate a list of superclasses for the class we're defining. For the purposes of this chapter, they're interchangeable (I don't think I ever inherit from anything).

Generic functions can specialize on all of them; `class`es, `struct`s and primitive types like `integer`/`list`/`character`/`string`/etc.

I'm not sure how much of this stuff is worth expanding on inline; but it might be a good idea to refer to more in-depth treatment (Chapters [16](http://www.gigamonkeys.com/book/object-reorientation-generic-functions.html) and [17](http://www.gigamonkeys.com/book/object-reorientation-classes.html) of PCL, for example).

event-web-framework/draft.markdown
Inaimathi: Should be ok for our purposes. Just two thoughts.

1. flipped column/row (so if we change the table, we should make the equivalent change in the acoompanying prose)
2. this is meant to contrast class-based and generic-based object systems. I'm not sure what multiple-dispatch would look like in something like Ruby or Java, so I don't know whether the modified version illustrates it accurately.

event-web-framework/draft.markdown
Inaimathi: The CL equivalent of `var`. It starts a variable definition form. The first symbol following it will be the name of the variable, and the second will be its value.

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+				,res)))
+		  ((list* arg-sym type restrictions)
+		   (setf res
+			 `(let ((,arg-sym ,(or (type-expression (arg-exp arg-sym) type restrictions) (arg-exp arg-sym))))
+			    ,@(awhen (type-assertion arg-sym type restrictions) `((assert-http ,it)))
+			    ,res))))
+	     finally (return res)))
+```
+
+Welcome to the hard part. `arguments` turns the validators we registered with our handler into a tree of parse attempts and assertions. `type-expression`, `arg-exp`, and `type-assertion` are used to implement and enforce a "type system" for the kinds of data we're expecting in our responses; we'll discuss them in SECTION FIXME. Using this together with `make-closing-handler`, this would the validation rules we wrote here:
```
Inaimathi: I assume that should read "Using this together with `make-closing-handler` would implement the validation rules we wrote here:"

Not sure the sentence makes sense anymore, given that this is the first place we introduce `define-handler` at this point. I think I'll need to finesse some of the prose based on these edits.

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+	     ,(arguments args
+			 `(let ((res (progn ,@body)))
+			    (write! (make-instance 'response
+						   :keep-alive? t :content-type "text/event-stream")
+				    sock)
+			    (write! (make-instance 'sse :data (or res "Listening...")) sock)
+			    (force-output (socket-stream sock))))))
+```
+
+	(defmacro assert-http (assertion)
```
Inaimathi: Fucked code-block here (the triple-dash should be below this `defmacro` form.

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+
+Recall that Lisp code is itself represented as a tree. That's what the parentheses are for; they show us how leaves and branches fit together. If we step back to what we were doing in the previous section, `make-closing-handler` calls a function called `arguments` to generate part of the Lisp tree its constructing, which in turn calls some tree-manipulating helper functions, including `arg-exp`, to generate its return value. 
+
+That is, we've built a small system that takes a Lisp expression as input, produces a different Lisp expression as output. Possibly the simplest way of conceptualizing this is as a simple Common Lisp to Common Lisp compiler that is specialized to the problem at hand.
+
+A widely used classification of such compilers are called _anaphoric macros_ [^anaphors]. This term comes from the linguistic concept of an _anaphor_, which is the use of one word as a substitute for a group of words that preceded it. `aif` and `awhen` are anaphoric macros, and they're the only ones that I tend to often use. There are many more availabile in the [`anaphora` package](http://www.cliki.net/Anaphora).
```
Inaimathi: Can't find the other end of this `[^anaphors]` anywhere. Did it get edited out, or was that unintentional?

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+
+This is a generic function that generates new tree structures (coincidentally Lisp code), rather than just a function. The only thing the above tells you is that by default, a `type-expression` is `NIL`. Which is to say, we don't have one. If we encounter a `NIL`, we use the raw output of `arg-exp`, but that doesn't tell us much about the most common case. To see that, lets take a look at a built-in (to `:house`) `define-http-type` expression.
+
+```lisp
+	(define-http-type (:integer)
+	    :type-expression `(parse-integer ,parameter :junk-allowed t)
+		:type-assertion `(numberp ,parameter))
+```
+
+An `:integer` is something we're making from a `parameter` by using `(parse-integer parameter :junk-allowed t)`. `junk-allowed` tells `parse-integer` that we're not confident the data we're giving it is actually parseable, so we need to make sure that the returned result is an integer. If it isn't, we get this behaviour:
```
Inaimathi: Note to self: Look at this and the accompanying REPL block. Might just be a brain-fart on my part, but it looks like this might not make sense.

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+	    `(let ((,tp ,type))
+	       ,@(when type-expression
+		       `((defmethod type-expression (parameter (type (eql ,tp)))
+			   ,type-expression)))
+	       ,@(when type-assertion
+		       `((defmethod type-assertion (parameter (type (eql ,tp)))
+			   ,type-assertion))))))
+```
+
+[^readable]: This macro is difficult to read because it tries hard to make its output human-readable, by expanding `NIL`s away using `,@` where possible.
```
Inaimathi: May want to expand on this by showing an expansion or two, given that this is already a footnote.

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+
+It works by creating `type-expression` and `type-assertion` method definitions for the type being defined. We could let users of our framewokr do this manually without much trouble; however, adding this extra level of indirection gives us, the framework programmers, the freedom to change _how_ types are implemented without forcing our users to re-write their specifications. This isn't just an academic consideration; I've personally made radical changes to this part of the system when first building it, and was pleased to find that I had to make very few edits to the applications that depended on it.
```
Inaimathi: framewokr -> framework

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+         (:html
+           (:head (:script :type "text/javascript" :src "/static/js/interface.js"))
+           (:body (:div :id "messages")
+	              (:textarea :id "input")
+	              (:button :id "send" "Send")))))
+
+    (start 4242)
+```
+
+Once we write `interface.js` to provide the client-side interactivity, this will start an HTTP chat server on port `4242` and listen for incoming connections. <some inspiring sentence>
```
Inaimathi: TODO - write an inspiring sentence

event-web-framework/draft.markdown
MichaelDiBernardo: It's close enough to how we explain it in our Julia chapter that I think we should be ok. Alternately I'm thinking about skipping the table analogy entirely, since it is also made in the Julia chapter in more detail that we'll be using here.

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,872 @@
...
+
+This means that in a class-based system, it's difficult to know what classes implement method `foo`; in a generic-function-based system, it is difficult to know all the methods that specialize on type `bar`. Depending on the nature of your system and your problem, you may want one or the other. (FIXME this probably needs more depth. Or just say how to make this decision is outside the scope of this chapter, we'll just give you a taste of what it's like to work in this type of system here.)
+
+### Processing sockets
+
+We'll see another generic function at work in `process-ready`, which was called earlier from our event loop. It processes a ready socket with one of two methods, depending on the type of socket we are handling.
+
+FIXME: What precisely is the difference between `stream-usocket` and `stream-server-usocket`? What coerces our sockets to one of these two types in the first place?
```
MichaelDiBernardo: Forgot to highlight this note -- @Inaimathi can you explain this a bit more?

event-web-framework/draft.markdown
MichaelDiBernardo: Awesome, those references are perfect. Thanks. (I ended up reading a few chapters of [this book that I had lying around](http://www.amazon.com/Lisp-3rd-Edition-Patrick-Winston/dp/0201083191) to give me some context, but it wasn't particularly newbie-friendly.)

event-web-framework/draft.markdown
MichaelDiBernardo: Cool, thanks.

event-web-framework/draft.markdown
MichaelDiBernardo: Sorry, 85 is what we have to work with :/ Let's get the content editorial stuff out of the way first, and then I'll make my edits based on what we work out here. Finally, you can open one last PR with just the code-alignment changes after all that's done.

event-web-framework/draft.markdown
MichaelDiBernardo: Thanks. I think it's ok to "forward-declare" some of these things, since we are hinting that they're going to be explained later.

event-web-framework/draft.markdown
MichaelDiBernardo: Oops. Will fix.

event-web-framework/draft.markdown
MichaelDiBernardo: Oops. That's vestigial, I had this inline stuff as a footnote at one point.

event-web-framework/draft.markdown
MichaelDiBernardo: :/

MichaelDiBernardo:
@Inaimathi if you can supply some hints to the last few "note" comments that have been left unaddressed so far, I can go back and do a second pass of edits which we can review here. Then we can look at having you do a final PR with code alignment changes + any content edits you want to try to "smooth out" the organizational changes I've made.

event-web-framework/draft.markdown
MichaelDiBernardo: I decided to just skip the table entirely.

Inaimathi: Tried to merge the 500 Lines changes back into my fork so I could open
another pull request, but the push is taking an agonizingly long time :/

Have some text files instead.

The markdown is an indented and modified copy of the chapter (with some
notes handled). The asd, and lisps are some minor code changes I had to
make to avoid blocking on writes (thanks for pointing that out
incidentally).

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,941 @@
...
+
+Our `buffer` class has six slots:
+
+- `tries`, which keeps count of how many times we've tried reading into this buffer
+- `contents`, which contains what we've read so far
+- `bi-stream`, which a hack around some of those Common Lisp-specific, non-blocking-IO annoyances I mentioned earlier
+- `total-buffered`, which is a count of chars we've read so far
+- `started`, which is a timestamp that tells us when we created this buffer
+- `request`, which will eventually contain the request we construct from buffered data
+- `expecting`, which will signal how many more chars we're expecting (if any) after we buffer the request headers
```
MichaelDiBernardo: I ended up not rendering these as tables, because bulleted list was easier to read in both web and print/PDF formats.

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,941 @@
...
+			 sock)
+		  collect it))))
+```
+
+In `publish!`, we call `write!` to actually write an `sse` to a socket. We'll also need a specialization of `write!` on `response`s to write full HTTP responses as well. Let's handle the HTTP case first.
+
+```lisp
+(defmethod write! ((res response) (socket usocket))
+  (handler-case
+      (with-timeout (.2)
```
MichaelDiBernardo: Is this spawning another thread under the hood?

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,941 @@
...
+	    (write-ln "Cache-Control: no-cache, no-store, must-revalidate")
+	    (when (keep-alive? res)
+	      (write-ln "Connection: keep-alive")
+	      (write-ln "Expires: Thu, 01 Jan 1970 00:00:01 GMT"))
+	    (awhen (body res)
+	      (write-ln "Content-Length: " (write-to-string (length it)))
+	      (crlf stream)
+	      (write-ln it))
+	    (values))))
+    (trivial-timeout:timeout-error ()
```
MichaelDiBernardo: Is it possible that we've partially written to this socket, and that we'll try to rewrite the whole response again from the beginning in another run through the loop?

event-web-framework/draft.markdown
```diff
@@ -0,0 +1,941 @@
...
+```lisp
+(defmethod write! ((res sse) (socket usocket))
+  (let ((stream (flex-stream socket)))
+    (handler-case
+    (with-timeout (.2)
+      (format
+       stream "~@[id: ~a~%~]~@[event: ~a~%~]~@[retry: ~a~%~]data: ~a~%~%"
+       (id res) (event res) (retry res) (data res)))
+      (trivial-timeout:timeout-error ()
+        (values)))))
```
MichaelDiBernardo: Same comment re: partial writes that will be fully retried here.

MichaelDiBernardo:
@Inaimathi thanks for the edits, this looks great. I just have a couple of last questions around the changes that were made re: writing responses to sockets without blocking that I'd like to have cleared up before we merge and go to copy-editing, because depending on the answers I may want to add a few more lines of explanation / justification.

event-web-framework/draft.markdown
Inaimathi: :+1: Thoroughly approve of lists supplanting tables.

event-web-framework/draft.markdown
Inaimathi: Implementation dependant. You can see the details at https://github.com/gwkkwg/trivial-timeout/blob/master/dev/with-timeout.lisp#L16-L120

(Some implementations provide a primitive timeout operation, which seems to be used where available. Others don't, and in those, it seems to spawn either a thread or a process.)

I'm not a hundred percent sure this is something we need to worry about in the first place; Common Lisp `write-sequence` doesn't seem to block on a foreign socket, even if data isn't read out of it. Maybe we just note that instead of including a separate external library?)

event-web-framework/draft.markdown
Inaimathi: Nope. If there's an error during a request `write!` we throw that socket out (we try to write an `error` response first, but at that point we just drop the client). Not entirely sure that's the *right* thing, but I'm not sure what else we could do if retries have made it up past TCP level.

event-web-framework/draft.markdown
MichaelDiBernardo: I wonder if we should do both -- we can mention here that blocking is less likely on a write, since it's likely being buffered already, but if the event that it does hang we don't want it to stop the entire server, hence the timeout. (People who want to learn to write server guts probably need to get used to seeing ugly timeout code anyways :) ) I can make that wording change now.

MichaelDiBernardo:
OK, I think we've managed to converge to something copy-editable. I'm going to copy `draft.md` over to `event-web-framework.md` in a subsequent commit, merge this, and get copy-editing started. Thanks @Inaimathi!

PR #199, opened 2016-03-04
Copyedits on Same-Origin Policy in Alloy chapter

amyrbrown:
Here are my edits for this chapter. There are a few comments on the Files Changed page, but they're mainly typesetting notes for @MichaelDiBernardo. 

Also:
On line 262, perhaps the electronic versions of the chapter could link to the Appendix here?
On line 1055 there is a footnote to be typeset.

And I wonder if it would be worthwhile to set the Appendix in slightly smaller text, to set it off from the main text of the chapter? Depending how much work it would be. :)

### DIFF
--- a/same-origin-policy/same-origin-policy.markdown
+++ b/same-origin-policy/same-origin-policy.markdown
@@ -1,11 +1,11 @@
-title: The Same Origin Policy
+title: The Same-Origin Policy
 author: Eunsuk Kang, Santiago Perez De Rosso, and Daniel Jackson
 
 _Eunsuk Kang is a PhD candidate and a member of the Software Design Group at MIT. He received his SM (Master of Science) in Computer Science from MIT (2010), and a Bachelor of Software Engineering from the University of Waterloo (2007). His research projects have focused on developing tools and techniques for software modeling and verification, with applications to security and safety-critical systems._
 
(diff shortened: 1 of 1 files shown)


same-origin-policy/same-origin-policy.markdown
```diff
@@ -745,7 +745,7 @@ assert Confidentiality {
 }
 ```
 
-The integrity property is the dual of confidentiality: 
+The integrity property is the double of confidentiality:
```
amyrbrown: I wasn't entirely sure what was meant here, but if you mean "the twin of" or "the integrity property is just like confidentiality", then "double" works better than "dual".

same-origin-policy/same-origin-policy.markdown
```diff
@@ -1222,13 +1222,13 @@ inject bad content as part of a `PostMessage` [cite PostMessage
 study].
```
amyrbrown: Flagging that we are still missing this citation.

same-origin-policy/same-origin-policy.markdown
```diff
@@ -1308,7 +1308,7 @@ Why would a developer ever use the wildcard? It turns out that
 specifying the allowed origins can be tricky, since it may not be
```
amyrbrown: There's a citation reference at line 1305 that's still missing.

same-origin-policy/same-origin-policy.markdown
eskang: In this context, we mean "integrity" is a concept that mirrors "confidentiality", so "dual" is more appropriate.

MichaelDiBernardo:
Merging this for now -- @eskang et al will make changes in a subsequent PR, and I'll finish things up after that re: any links / refs I missed the first time around.

Issue #198, opened 2016-03-03
Assertion failure in cluster.py


1-p:
if you run with `python run.py 903` you may encounter an assertion failure.

    AssertionError: next slot to commit is already decided

MichaelDiBernardo:
@djmitche, are you available to comment on this? I can do it eventually, but it's going to be closer to the end of March.

Thanks for the report @1-p.

1-p:
@MichaelDiBernardo the output is here https://gist.github.com/1-p/01ff5fe68e81c7d11bcf

(I added some code to print the internal state at the end when assertion failed.)

Somehow `do_Decision` is called with `slot = 27` when decisions contains 0-29.

djmitche:
`slot` is the next slot that the replica wants to commit, but does not yet have a decision for.  The `while` loop in `do_Decision` should take care of ensuring this invariant, by looping until `self.decisions.get(self.slot)` is false.  In developing the code, I've seen this happen when `decisions` dictionaries are aliased between components -- that is, when two different nodes are using the same Python dictionary.  Then one node adds key `27` to the dictionary and increments its own `slot`, and another component comes along and is surprised to see `27` in the dictionary.

So I see N2 welcome N0 in a state where slots 1-26 are decided, and slot is 27.  Everything still makes sense here
```
[...]

PR #200, opened 2016-03-05
Fix data aliasing issues

djmitche:
This categorically addresses the unintended sharing of data structures between cluster components; #198 is one example of this issue.

Note that this makes a fresh copy of the message for each destination, avoiding aliasing either between the source and destination, or between multiple destinations.

This requires a little trickiness in Python: since we use `message` in a closure, we need a distinct variable for each destination, and we need to make that copy during the `send` invocation, not when the timer expires.

### DIFF
--- a/cluster/code/cluster.py
+++ b/cluster/code/cluster.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+import copy
 import functools
 import heapq
 import itertools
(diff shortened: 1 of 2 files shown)


MichaelDiBernardo:
Cool, thank you @djmitche! I'll check up before publication to make sure that these code changes propagate to anyplace they may be in the chapter.

djmitche:
Awesome, thanks!

PR #201, opened 2016-03-06
removed ending bracket

dxnn:
Just removing an errant ending bracket from a footnote.


MichaelDiBernardo:
I hate it when I do that. Thanks :)

PR #223, opened 2016-05-31
New images for title, subtitle and editor in frontmatter. 

amyrbrown:
These files are probably the wrong size so there will need to be some futzing with the LaTeX to get them to look okay.


MichaelDiBernardo:
Thank you!

PR #231, opened 2016-06-23
New, bigger versions of editor and title images for frontmatter.

amyrbrown:
Hopefully these will look better. Let me know if they're too big (pixel-wise) and I'll resize them. (I can't remember if that can be effectively done in LaTeX.)


MichaelDiBernardo:
Thanks! :)

PR #234, opened 2016-07-21
Removed the '' at the beginning of the sed command

r-marques:
The extra `''` at the beginning of the sed command was generating these errors:

``` bash
./_build/fix_html_title.sh html/content/pages/web-server.md

sed: can't read s|<p>\(title:.*\)author:\(.*\)</p>|\1\
author:\2\
|: No such file or directory
```

And

``` bash
ERROR: Skipping pages/web-server.md: could not find information about 'NameError: title'
```

for all `md` files


chriscz:
Could someone please merge this fix? As it causes the html generation to fail on my system as well.

PR #269, opened 2017-11-02
:bug: fix mixing URL and string in crawler

Motor-taxi-master-rider:
This will raise an error in python 3.6

### DIFF
--- a/crawler/code/crawling.py
+++ b/crawler/code/crawling.py
@@ -121,13 +121,12 @@ class Crawler:
         """Record the FetchStatistic for completed / failed URL."""
         self.done.append(fetch_statistic)
 
-    @asyncio.coroutine
-    def parse_links(self, response):
+    async def parse_links(self, response):
         """Return a FetchStatistic and list of links."""
         links = set()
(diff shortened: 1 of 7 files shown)


crawler/code/crawling.py
```diff
@@ -147,7 +147,7 @@ def parse_links(self, response):
                     LOGGER.info('got %r distinct urls from %r',
                                 len(urls), response.url)
                 for url in urls:
-                    normalized = urllib.parse.urljoin(response.url, url)
+                    normalized = urllib.parse.urljoin(str(response.url), url)
```
asvetlov, line 149: `response.url.join(url)` is even better.

crawler/code/supplemental/loop-with-coroutines.py
```diff
@@ -152,7 +152,7 @@ def _is_html(self):
 start = time.time()
 fetcher = Fetcher('/')
 Task(fetcher.fetch())
-
+2
```
asvetlov, line 155: What?

PR #278, opened 2019-03-02
Fix typo

RanceJen:


utensil:
It dosen't seem to be a typo. The original text meant the attributes of messages. The modified version means the attributes of one message and is syntactically incomplete.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

RanceJen:
Thank you for the explanation

